PR #1, opened 2016-04-14
Adding a PumpManager class

mtlynch:
Adds a PumpManager class with unit tests. One unit test works, two fail, and
one is not implemented yet.


greenpithumb/pump_manager.py
```diff
@@ -0,0 +1,36 @@
...
+        """Run the water pump if there is a need to run it.
+
+        Checks whether it's time to run the pump and runs it if needed.
+        Running the pump is necessary if the soil's moisture is below the
+        defined moisture threshold.
+
+        """
+        # TODO(JeetShetty): Add logic here so that it only runs the pump when
+        # the soil moisture is too low.
+        self._pump.pump_water(DEFAULT_PUMP_AMOUNT)
```
JeetShetty: How can we do this without having defined a pump_water method?

tests/test_pump_manager.py
```diff
@@ -0,0 +1,43 @@
+from __future__ import absolute_import
+import unittest
+
+import mock
+
+from greenpithumb import pump_manager
+
+
+class PumpManagerTest(unittest.TestCase):
+
```
JeetShetty: This is pretty foreign/confusing to me and I'm having a tough time debugging and figuring out why my change is failing the "moisture_is_low" case as a result of that.

greenpithumb/pump_manager.py
JeetShetty: if self._moisture_sensor < SOIL_MOISTURE_THRESHOLD:
            self._pump.pump_water(DEFAULT_PUMP_AMOUNT)

mtlynch: It's missing a method. `self._moisture_sensor` is an object, so we need to call its `.moisture()` method to get the moisture, but once you do that it should work.

Also if you surround your code with triple backticks (\`\`\`), it formats nicely on Github comments

PR #7, opened 2016-04-24
Throwaway

JeetShetty:


tests/test_light_sensor.py
```diff
@@ -0,0 +1,39 @@
...
+		
+		self.mock_adc.read_pin.return_value = (
+			(light_sensor._LIGHT_SENSOR_MAX_VALUE - \
+			light_sensor._LIGHT_SENSOR_MIN_VALUE) / 2) + (
+			light_sensor._LIGHT_SENSOR_MIN_VALUE)
+		light_level = self.light_sensor.get_light_level()
+		self.assertAlmostEqual(light_level, 50.0)
+
+
+	def test_light_out_of_range(self):
```
mtlynch: I'd make these two separate tests.

tests/test_light_sensor.py
```diff
@@ -0,0 +1,39 @@
...
+		#ValueError should be raised if light level returned
+		#is either higher than max value or lower than min value
```
mtlynch: Comments documenting a function overall should generally be in triple quotes.

See:

https://www.python.org/dev/peps/pep-0257/#one-line-docstrings

Docstrings have special utility in unit test functions because the one-line summary in a docstring is what gets printed out when the test fails.

tests/test_light_sensor.py
```diff
@@ -0,0 +1,39 @@
...
+
+import mock
+
+from greenpithumb import adc
+from greenpithumb import light_sensor
+
+class LightSensorTest(unittest.TestCase):
+
+	def setUp(self):
+		self.mock_adc = mock.Mock(spec=adc.Adc)
```
mtlynch: Good mockery!

tests/test_light_sensor.py
```diff
@@ -0,0 +1,39 @@
...
+		self.light_sensor = light_sensor.LightSensor(self.mock_adc)
+
+	def test_light_50_pct(self):
+		#light sensor value midway between max and min should 
+		#return 50.0
+		
+		self.mock_adc.read_pin.return_value = (
+			(light_sensor._LIGHT_SENSOR_MAX_VALUE - \
+			light_sensor._LIGHT_SENSOR_MIN_VALUE) / 2) + (
+			light_sensor._LIGHT_SENSOR_MIN_VALUE)
```
mtlynch: It's a bit of a red flag when the unit test code reimplements logic from the production code. This is pretty similar to these lines:

https://github.com/JeetShetty/GreenPiThumb/blob/throwaway/greenpithumb/light_sensor.py#L27-28

I suggest just hardcoding the value in the test, like:

```
# Midpoint between LIGHT_SENSOR_MIN_VALUE and LIGHT_SENSOR_MAX_VALUE
self.mock_adc.read_pin.return_value = 656.5
```

It'll break if we have to change the constants, but it reduces the chances of us missing something because we've made the same error in coding logic in both places.

mtlynch:
LGTM

PR #8, opened 2016-04-30
add HumiditySensor class

JeetShetty:


greenpithumb/dht11_exceptions.py
```diff
@@ -0,0 +1,10 @@
+#Custom exceptions for DHT11 sensor
+
+class Error(Exception):
+	pass
+
+class MissingDataError(Error):
+	pass
+
+class IncorrectCRCError(Error):
+	pass
```
mtlynch: Oh we actually want these in `humidity_sensor.py`

greenpithumb/humidity_sensor.py
```diff
@@ -0,0 +1,27 @@
...
+		Args:
+			dht11_result: Result of a reading from a DHT11 humidity and
+			temperature sensor.   
+		"""
+		self._dht11_result = dht11_result
+
+	def get_humidity_level(self):
+		"""Returns humidity level."""
+
+		if self._dht11_result.error_code() == 1:
```
mtlynch: Once we import the DHT11 library, we'll want to replace these with the constants the library exports. Can we add a comment like

```
# TODO(JeetShetty): Replace these with constants from dht11 module
```

For fun, you can also create an [issue](https://github.com/JeetShetty/GreenPiThumb/issues) if you want to see what that's like.

greenpithumb/humidity_sensor.py
```diff
@@ -0,0 +1,27 @@
...
+			raise dht11_exceptions.MissingDataError()
+		elif self._dht11_result.error_code() == 2:
+			raise dht11_exceptions.IncorrectCRCError()
+		elif self._dht11_result.error_code() == 0:
+			humidity_level = self._dht11_result.humidity()
```
mtlynch: This is a bit subtle because it almost makes `0` seem like a failing error code. It's not super obvious that this is the success case.

I generally like to structure error handling so that I check for errors first, then after error checking, proceed like normal, so this would be like

```python
if self._dht11_result.error_code() != 0:
  if if self._dht11_result.error_code() == 1:
    ...
  elif self._dht11_result.error_code() == 2:
    ..
  else:
    raise ValueError("DHT11 error code out of range")

return humidity_level
```

greenpithumb/humidity_sensor.py
```diff
@@ -0,0 +1,27 @@
...
+		else:
+			raise ValueError("DHT11 error code out of range")
```
mtlynch: Similar to the other PR, it might be helpful to include the unexpected error code in the error message.

greenpithumb/humidity_sensor.py
```diff
@@ -0,0 +1,27 @@
...
+			dht11_result: Result of a reading from a DHT11 humidity and
+			temperature sensor.   
+		"""
+		self._dht11_result = dht11_result
+
+	def get_humidity_level(self):
+		"""Returns humidity level."""
+
+		if self._dht11_result.error_code() == 1:
+			raise dht11_exceptions.MissingDataError()
```
mtlynch: For this and for `IncorrectCRCError` we should include human-readable messages like

```python
raise MissingDataError('DHT11 sensor reported missing data')
```

greenpithumb/humidity_sensor.py
```diff
@@ -0,0 +1,27 @@
+import dht11_exceptions
+
+class HumiditySensor(object):
+	"""Wrapper for humidity sensor."""
+
+	def __init__(self, dht11_result):
```
mtlynch: I think it makes more sense for `HumiditySensor`'s contructor to take as  a parameter an instance of `dht11.DHT11` (similar to how the other classes take an `adc` instance). Then `HumiditySensor` is responsible for calling the instance's `read()` function and also handling the result.

tests/test_humidity_sensor.py
```diff
@@ -0,0 +1,44 @@
...
+
+	def setUp(self):
+		self.mock_dht11_result = mock.Mock()
+		self.humidity_sensor = humidity_sensor.HumiditySensor(
+			self.mock_dht11_result)
+
+	def test_humidity_50(self):
+		"""Value given should be value returned."""
+
+		self.mock_dht11_result.error_code.return_value = 0
```
mtlynch: I think this might be a moo point if we change to taking a `DHT11` instance instead of a result, but for mocking simple attributes (as opposed to functions), I think we can just do

```python
self.mock_dht11_result.error_code = 0
```

greenpithumb/humidity_sensor.py
```diff
@@ -0,0 +1,35 @@
+import dht11_exceptions
+
+class HumiditySensor(object):
+	"""Wrapper for humidity sensor."""
```
mtlynch: Suggest mentioning that it wraps the DHT11 sensor.

greenpithumb/humidity_sensor.py
```diff
@@ -0,0 +1,35 @@
...
+
+		Args:
+			dht11: Instance of DHT11 sensor reader class. 
+		"""
+		self._dht11 = dht11
+
+	def get_humidity_level(self):
+		"""Returns humidity level."""
+
+		error_code = self._dht11.read().error_code
```
mtlynch: We want to hold onto the result after we read from dht11. This and line 33 are two separate reads, so if we find a zero error code here, the error code for 35's read might not be zero.

So we'd want to save it with something like

```python
dht11_result = self._dht11.read()
```

tests/test_humidity_sensor.py
```diff
@@ -0,0 +1,44 @@
...
+class HumiditySensorTest(unittest.TestCase):
+
+	def setUp(self):
+		self.mock_dht11 = mock.Mock()
+		self.humidity_sensor = humidity_sensor.HumiditySensor(
+			self.mock_dht11)
+
+	def test_humidity_50(self):
+		"""Value given should be value returned."""
+
```
mtlynch: We generally shouldn't start methods with blank lines. I use blank lines in the places where the style guide mandates it, but other than that I try to limit blank lines so that I only use them to logically group blocks of code.

>Two blank lines between top-level definitions, be they function or class definitions. One blank line between method definitions and between the class line and the first method. Use single blank lines as you judge appropriate within functions or methods.

https://google.github.io/styleguide/pyguide.html?showone=Blank_Lines#Blank_Lines

tests/test_humidity_sensor.py
```diff
@@ -0,0 +1,44 @@
...
+		self.mock_dht11.read().error_code = 0
+		self.mock_dht11.read().humidity = 50.0
```
mtlynch: I'm actually kind of surprised this works. We might be relying on undefined behavior, meaning that it could change by chance in a minor update to Python or it might have different behavior on different systems.

What I'd expect to happen is that each `read()` creates a new, independent object that gets thrown away at the end of the line. But the `mock` library does a lot of crazy magic that I think makes this work for whatever reason right now, but we probably shouldn't depend on it.

I'd actually suggest fixing this in `setUp` so instead of keeping around `mock_dht11`, we keep around the result, so like:

```python
def setUp(self):
  self.mock_dht11_result = mock.Mock()
  mock_dht11 = mock.Mock(return_value=self.mock_dht11_result)
  self.humidity_sensor = humidity_sensor.HumiditySensor(mock_dht11)

  # The last two lines can actually reduce to:
  self.humidity_sensor = humidity_sensor.HumiditySensor(mock.Mock(return_value=self.mock_dht11_result))
```

Then in this test, we could do:

```python
def test_humidity_50(self):
  self.mock_dht11_result.error_code = 0
  self.mock_dht11_result.humidity = 50.0
  ...
```

Notice that we don't need to hang onto a reference to `mock_dht11` by assigning it to `self` because everything we need to manipulate can happen through `self.mock_dht11_result`.

JeetShetty: I'm having trouble getting this to work. It appears that with the way you have it set up above the error code and humidity aren't getting assigned values.

I'm also a little confused about what you mean when you say "What I'd expect to happen is that each read() creates a new, independent object that gets thrown away at the end of the line" and why this is an issue.

tests/test_humidity_sensor.py
```diff
@@ -0,0 +1,40 @@
...
+from greenpithumb import humidity_sensor
+from greenpithumb import dht11_exceptions
+
+
+class HumiditySensorTest(unittest.TestCase):
+
+    def setUp(self):
+        self.mock_dht11_result = mock.Mock()
+        self.humidity_sensor = humidity_sensor.HumiditySensor(mock.Mock(
+            return_value=self.mock_dht11_result))
```
mtlynch: >I'm having trouble getting this to work. It appears that with the way you have it set up above the error code and humidity aren't getting assigned values

Whoops, I got this wrong. We want `self.mock_dht11_result` to be the return value from the `read` method, not the return value from the object itself. Created a PR you can pull into this branch to fix it:

https://github.com/JeetShetty/GreenPiThumb/pull/16

PR #9, opened 2016-04-30
add moisture sensor

JeetShetty:


greenpithumb/moisture_sensor.py
```diff
@@ -1,5 +1,31 @@
+import adc
+
+#placeholder values
+_MIN_MOISTURE_LEVEL = 0.0
+_MAX_MOISTURE_LEVEL = 100.0
```
mtlynch: I think the range is 0 to 1023, based on this
https://github.com/dicksondickson/PlantFriends/blob/master/sensor_node/RFM69/sensor_node.ino#L23-31

greenpithumb/moisture_sensor.py
```diff
@@ -1,5 +1,31 @@
...
+
+    	Args:
+    		adc: ADC(analog to digital) interface to receive 
+    		analog signals from moisture sensor.
+    	"""
+    	self._adc = adc
+
     def moisture(self):
-        raise NotImplementedError()
+        """Returns moisture level as percent."""
```
mtlynch: Do you think it makes sense to convert it to a percentage? It doesn't look like it scales linearly, so it might be better just to leave it as a value of arbitrary scale.

mtlynch: We need to update the comment since it's no longer a percentage.

greenpithumb/moisture_sensor.py
```diff
@@ -1,5 +1,28 @@
...
+
+        moisture_level = self._adc.read_pin(adc.PIN_MOISTURE_SENSOR)
+
+        if moisture_level < _MIN_MOISTURE_LEVEL or \
+        moisture_level > _MAX_MOISTURE_LEVEL:
+        	raise ValueError('moisture level out of range')
```
mtlynch: Might be helpful to include the value too, like

```
raise ValueError('moisture level out of range: %.1f' % moisture_level)
```

tests/test_moisture_sensor.py
```diff
@@ -0,0 +1,37 @@
...
+	def test_moisture_50(self):
+		"""Moisture level read should be moisture level returned."""
+
+		self.mock_adc.read_pin.return_value = 50.0 
+		moisture_level = self.moisture_sensor.moisture()
+		self.assertEqual(moisture_level, 50.0)
+
+
+	def test_moisture_too_low(self):
+		"""moisture lower than min level should raise a ValueError."""
```
mtlynch: For this, consider just explaining in the test name and dropping the docstring like

```
test_moisture_too_low_raises_ValueError(self)
```

mtlynch:
A few notes, but LGTM

PR #10, opened 2016-05-06
Adding DHT11 submodule

mtlynch:
Adding the Python DHT11 library as a submodule and adding a dependency on
RPi.GPIO since DHT11 needs it.


mtlynch:
Whoops, hang on, this is wrong

PR #15, opened 2016-05-10
add TemperatureSensor class

JeetShetty:


greenpithumb/temperature_sensor.py
```diff
@@ -0,0 +1,36 @@
...
+
+class TemperatureSensor(object):
+    """Wrapper for temperature sensor."""
+
+    def __init__(self, dht11):
+        """Creates a new TemperatureSensor wrapper.
+
+		Args:
+			dht11: Instance of DHT11 sensor reader class. 
+		"""
```
mtlynch: Weird spacing going on here. Looks like we might be mixing tabs and spaces.

mtlynch:
Looks like basically same notes I had in #8 apply here as well

greenpithumb/temperature_sensor.py
```diff
@@ -0,0 +1,33 @@
...
+    def __init__(self, dht11):
+        """Creates a new TemperatureSensor wrapper.
+
+        Args:
+            dht11: Instance of DHT11 sensor reader class. 
+        """
+        self._dht11 = dht11
+
+    def get_temperature(self):
+        """Returns ambient temperature."""
```
mtlynch: We should say in what units. I believe it's in celsius.

mtlynch:
LGTM

PR #17, opened 2016-05-14
Update pump_manager to take time of day into account

JeetShetty:


mtlynch:
Cool! So my major note is that I don't think the waking time and sleeping time belong in `clock.py` because those concepts are probably specific to `PumpManager` and we want `Clock` to be a general purpose clock that everything uses.

We probably also don't want to hardcode the bedtime hour and waketime hour because you might want to go to sleep at 10pm, but I might want to go to sleep at 11pm, so we can make it configurable. I can think of two options:

1 - Make everything a parameter to `PumpManager`'s `__init__` like:

```python
def __init__(self, moisture_sensor, pump, clock, bedtime_hour, waketime_hour):
```

This is simple to implement, but it's a bit inflexible because if we decide we want to allow minute-specificity or multiple time windows to sleep (like don't turn on when I'm sleeping or when I'm eating dinner), we'd have to add too many parameters.

That said, this would be fine for what we need right now.

2 - Make a `PumpScheduler` class like

```python
class PumpScheduler(object):

  def __init__(self, clock, sleep_hour_local, wake_hour_local):
    self._clock = clock
    # TO JEET: I'm including the _local suffix to make it explicit it's in local time, even though
    # clock gives time in UTC, so we'll need to convert time zones.
    self._sleep_hour_local = sleep_hour_local
    self._wake_hour_local = wake_hour_local

  def run_pump_ees_okay(self):
    ...
```

Or maybe we call the function something like `is_running_pump_allowed`.

But then that way we put all the scheduling logic in `PumpScheduler` and even if we make the scheduling more complicated, nothing changes about `PumpManager` because it just takes a `PumpScheduler` instance and only runs if `is_running_pump_allowed` is `True`.

JeetShetty:
Hey so I've been perusing datetime and pytz and I'm having trouble figuring out how we figure out what the local time zone is? Also it seems to make more sense to me to have clock.now() take the local timezone as a second parameter so that everything is returned in local time.

mtlynch:
>Hey so I've been perusing datetime and pytz and I'm having trouble figuring out how we figure out what the local time zone is?

Yeah, Python's really annoying about time zones. I think we need to add `tzlocal` to `requirements.txt` and then do something like this:

```python
import tzlocal
...

time_utc = self._clock.now()
time_local = time_utc.astimezone(tzlocal.get_localzone())
```

>Also it seems to make more sense to me to have clock.now() return the time in the local time zone

If you want to, we can because it probably won't matter that much for a project this narrowly scoped, but in general it's good practice to use UTC as the base time and only convert to local when we need to.

There's some weirdness with using local time, like due to daylight saving time, the time 1 AM occurs twice on Nov. 6, 2016, whereas with UTC it doesn't. It can lead to subtle bugs, but a lot of them aren't that likely here.

If we want local time, we'd drop `pytz` from requirements and change the call to:

```python
return datetime.datetime.now(tz=tzlocal.get_localzone())
```

PR #20, opened 2016-05-20
Adding a basic command-line interface

mtlynch:
Adds a basic command line interface for GreenPiThumb. At this point, all it
does is read each sensor and print out the value.

Since we're not running on real hardware yet, adds a fake_sensors module that
generates random sensor values.

### DIFF
--- a/build
+++ b/build
@@ -1,8 +1,7 @@
 # Run unit tests and calculate code coverage.
 coverage run \
   --source greenpithumb \
-  --omit "greenpithumb/dht11/*" \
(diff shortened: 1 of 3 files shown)


mtlynch:
Whoops, I messed up the build file. Fixing...

mtlynch:
Okay, now it's fixed. The code coverage dropped a lot because the file that implements the command-line interface is not under test (it's hard to test the CLI and in this case it's not that useful).

Our code coverage percent will go back up as we add more files and test them, though!

PR #21, opened 2016-05-22
Add PumpScheduler class.

JeetShetty:


greenpithumb/pump_scheduler.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def is_running_pump_allowed(self):
+        """Returns True if OK to run pump, otherwise False."""
+        time_utc = self._clock.now()
+        time_local = time_utc.astimezone(tzlocal.get_localzone())
+        hour = time_local.hour
+
+        for window in self._sleep_windows_local:
+            window[0] = sleep_hour
+            window[1] = wake_hour
```
mtlynch: This looks backwards since we want to assign the other way around. There's a neat way of doing this easily in python like:

```
for sleep_hour, wake_hour in self._sleep_windows_local:
```

greenpithumb/pump_scheduler.py
```diff
@@ -0,0 +1,36 @@
...
+
+            if wake_hour < sleep_hour:
+                if hour >= sleep_hour or hour < wake_hour:
```
mtlynch: We can also do

```
if sleep_hour <= hour < wake_hour:
```

tests/test_pump_scheduler.py
```diff
@@ -0,0 +1,47 @@
...
+        self.mock_clock = mock.Mock()
+
+    def test_wake_hour_less_than_sleep_hour(self):
+        """Test rollover from one day to the next.
+        
+        If the sleep period spans two days, the pump should not run between 
+        midnight and the wake hour.
+        """
+        self.mock_clock.now.return_value = datetime.datetime(2016, 5, 24, 0)
+        sleep_windows = ((22, 8),)
```
mtlynch: (minor, optional) This is correct, but looks kind of weird. Suggest changing it to a list (square brackets) just because a list is fine here too and the syntax looks a little clearer like:

```python
sleep_windows = ([22,8])
```

(ditto for the rest, just for consistency)

tests/test_pump_scheduler.py
```diff
@@ -0,0 +1,47 @@
...
+        mock_pump_scheduler = pump_scheduler.PumpScheduler(self.mock_clock,
```
mtlynch: The pump scheduler is real (it's a real instance of `PumpScheduler`), so it shouldn't have a `mock_` prefix.

The problem is that calling it `pump_scheduler` gives it the same name as the module. We can just not assign it to a variable at all and do

```python
self.assertFalse(pump_scheduler.PumpScheduler(self.mock_clock, sleep_windows).is_running_pump_allowed())
```

greenpithumb/pump_scheduler.py
```diff
@@ -0,0 +1,31 @@
...
+        self._sleep_windows = sleep_windows
+
+    def is_running_pump_allowed(self):
+        """Returns True if OK to run pump, otherwise False."""
+        time = self._clock.now()
+        hour = time.hour
+
+        for sleep_hour, wake_hour in self._sleep_windows:
+            if wake_hour < sleep_hour:
+                if hour >= sleep_hour or hour < wake_hour:
```
mtlynch: We can do the simplified syntax here too like

```python
if sleep_hour <= hour < wake_hour:
```

greenpithumb/pump_scheduler.py
```diff
@@ -0,0 +1,31 @@
...
+            sleep_windows: A tuple of tuples, each consisting of a sleep 
+                window. Time zone should match that of the clock interface.
+        """
+        self._clock = clock
+        self._sleep_windows = sleep_windows
+
+    def is_running_pump_allowed(self):
+        """Returns True if OK to run pump, otherwise False."""
+        time = self._clock.now()
+        hour = time.hour
```
mtlynch: Suggest simplifying to

```python
current_hour = self._clock.now().hour
```

tests/test_pump_scheduler.py
```diff
@@ -0,0 +1,47 @@
...
+                                                           sleep_windows)
+        self.assertFalse(mock_pump_scheduler.is_running_pump_allowed())
+
+    def test_running_pump_allowed_if_not_in_sleep_window(self):
+        self.mock_clock.now.return_value = datetime.datetime(2016, 5, 24, 10)
+        sleep_windows = ((2, 8),)
+        mock_pump_scheduler = pump_scheduler.PumpScheduler(self.mock_clock,
+                                                           sleep_windows)
+        self.assertTrue(mock_pump_scheduler.is_running_pump_allowed())
```
mtlynch: Suggest adding a test case for when the current hour is the sleep hour and and one where the current hour is the wake hour.

greenpithumb/pump_scheduler.py
JeetShetty: I don't think we can do this, because sleep_hour > hour if hour is midnight or after.

mtlynch: Oh yeah you're right. I was thinking `and` instead of `or`.

tests/test_pump_scheduler.py
mtlynch: Oh sorry I'm stupid and changed the wrong brackets.

```python
sleep_windows = ([22,8])
```

just evaluates to:

```python
sleep_windows = [22,8]
```

Because it thinks the parens are for precedence. We want the other way around

```python
sleep_windows = [(22, 8)]
```

greenpithumb/pump_scheduler.py
```diff
@@ -0,0 +1,30 @@
...
+    def is_running_pump_allowed(self):
+        """Returns True if OK to run pump, otherwise False."""
+        current_hour = self._clock.now().hour
+
+        for sleep_hour, wake_hour in self._sleep_windows:
+            if wake_hour < sleep_hour:
+                if current_hour >= sleep_hour or current_hour < wake_hour:
+                    return False
+            else:
+                if sleep_hour <= current_hour < wake_hour:
```
mtlynch: We can shorten the `else: ... if` to just `elif` since it's the only thing in the `else`

mtlynch:
A couple minor notes but LGTM

tests/test_pump_scheduler.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def test_wake_hour_less_than_sleep_hour(self):
+        """Test rollover from one day to the next.
+        
+        If the sleep period spans two days, the pump should not run between 
+        midnight and the wake hour.
+        """
+        self.mock_clock.now.return_value = datetime.datetime(2016, 5, 24, 0)
+        sleep_windows = ((22, 8),)
+        pump_sched = pump_scheduler.PumpScheduler(self.mock_clock,
```
mtlynch: I try to avoid abbreviations when I can. Suggest `scheduler`

PR #22, opened 2016-05-26
Adding docstring checker and pylint checks

mtlynch:
Adding two additional automated checks

* pyflakes will look for cruft like variables we declare but don't ever use or
  imports we make but don't use (it flagged one in pump.py)
* docstringchecker is a pylint plugin that will check to make sure all of
  our docstrings are properly formatted (it flagged two instances where we had
 trailing whitespace).

### DIFF
--- a/third_party/docstringchecker/lint.py
+++ b/third_party/docstringchecker/lint.py
@@ -1,0 +1,490 @@
+# Copyright (c) 2013 The Chromium OS Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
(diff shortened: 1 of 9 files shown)


third_party/docstringchecker/lint.py
```diff
@@ -0,0 +1,490 @@
+# Copyright (c) 2013 The Chromium OS Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
```
mtlynch: To be clear, this is third-party code we're pulling in. Ideally we wouldn't have to copy it locally, but there isn't a good way of getting it otherwise. You don't need to code review it, though.

PR #23, opened 2016-05-27
Weight sensor

JeetShetty:
This change is [Reviewable](https://reviewable.io/reviews/jeetshetty/greenpithumb/23)


mtlynch:
Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[greenpithumb/weight_sensor.py, line 3 \[r2\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/23#-KJNCbNAU_Dax7WuyDLF:-KJNC-Qlg9wOK3CqJ4ND:-1243628324) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/f2279a310268481a7e2a6592418d62538df1c9c2/greenpithumb/weight_sensor.py#L3)):*
> ```Python
> import adc
> 
> _CONTAINER_WEIGHT = 100.0  #placeholder
> ```

I think this should be a parameter to the `__init__` function since we might not have the same weight containers. This is another thing that should be in a config file eventually.

---

*[greenpithumb/weight_sensor.py, line 9 \[r2\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/23#-KJNEPccDFeGgwNDlybq:-KJNEPccDFeGgwNDlybr:1832924718) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/f2279a310268481a7e2a6592418d62538df1c9c2/greenpithumb/weight_sensor.py#L9)):*
> ```Python
> 
> 
> class WeightSensor(object):
> ```

This is a bit unintuitive because it's called `WeightSensor` but doesn't report weight. Right now, this class seems to be mixing functionality in a few ways.

* It interprets what the weight sensor is saying through the ADC
* It converts a weight to a volume measurement
* It decides whether a water volume is too low

According to [Curly's Law](https://blog.codinghorror.com/curlys-law-do-one-thing/) each class should just do one thing, so it makes sense to break this up.

We could do something like:

* `WeightSensor` - Determines weight from an ADC
* `Reservoir` - Determines reservoir water level, given a `WeightSensor`
* `ReservoirMonitor` - Monitors a `Reservoir` instance and triggers an alert (sensor light, email notification, some notice in the web interface) when reservoir level is too low

---

*[greenpithumb/weight_sensor.py, line 25 \[r2\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/23#-KJND_mv3QOw-ALZF2YH:-KJND_mv3QOw-ALZF2YI:1243016744) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/f2279a310268481a7e2a6592418d62538df1c9c2/greenpithumb/weight_sensor.py#L25)):*
> ```Python
>         reservoir_weight = self._adc.read_pin(adc.PIN_WEIGHT_SENSOR)
>         if not _MIN_RES_WEIGHT <= reservoir_weight <= _MAX_RES_WEIGHT:
>             raise ValueError('reservoir weight out of range: %.1F' %
> ```

Does uppercase F work? I've only ever seen like `%f`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/23)*

mtlynch:
Reviewed 2 of 3 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/23)*

JeetShetty:
*[greenpithumb/weight_sensor.py, line 25 \[r2\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/23#-KJND_mv3QOw-ALZF2YH:-KJS1yuACMAaaFfoe4iu:-362711764) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/f2279a310268481a7e2a6592418d62538df1c9c2/greenpithumb/weight_sensor.py#L25)):*
> Does uppercase F work? I've only ever seen like `%f`
</details>
uppercase F does work, but I didn't mean to do that! Will switch to lowercase

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/23)*

mtlynch:
Reviewed 5 of 5 files at r3.
Review status: all files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[greenpithumb/reservoir.py, line 5 \[r3\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/23#-KJXOMTaJ0jm-Oz3JATW:-KJXOMTaJ0jm-Oz3JATX:1500038218) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/efffc278d69dfdadaf1f527d6386cf3280d5bdea/greenpithumb/reservoir.py#L5)):*
> ```Python
> 
> class Reservoir(object):
>     """Wrapper for a reservoir."""
> ```

Wrapper probably isn't quite the right term here. Wrapper makes sense for things like the weight sensor because there actually is a physical sensor that gives out signals and we're wrapping it with a little bit of code so that there's a Python interface to represent the signals it gives off.

`Reservoir` is getting a bit more abstract because there's no "reservoir sensor" that we're calling to get this value. We're creating this concept by putting together other information. (ditto for `ReservoirMonitor`).

Suggestion:

```python
"""Provides information about a GreenPiThumb's system's water reservoir."""
```

---

*[greenpithumb/reservoir.py, line 19 \[r3\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/23#-KJXHBHch2ksl-o15SzC:-KJXHBHch2ksl-o15SzD:1801377056) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/efffc278d69dfdadaf1f527d6386cf3280d5bdea/greenpithumb/reservoir.py#L19)):*
> ```Python
>     def reservoir_level(self):
>         """Returns reservoir level in ml."""
>         res_weight_grams = self._weight_sensor.reservoir_weight()
> ```

Kind of weird to abbreviate to "res" here when we're using "reservoir" everywhere else

---

*[greenpithumb/reservoir.py, line 21 \[r3\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/23#-KJXHaFkXJaDiHakgbeH:-KJXHaFkXJaDiHakgbeI:-604671107) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/efffc278d69dfdadaf1f527d6386cf3280d5bdea/greenpithumb/reservoir.py#L21)):*
> ```Python
>         res_weight_grams = self._weight_sensor.reservoir_weight()
>         water_weight_grams = res_weight_grams - self._container_weight
>         if water_weight_grams < weight_sensor.MIN_RES_WEIGHT:
> ```

This shouldn't be necessary. It looks like `WeightSensor` will already throw an exception in this case so this condition should never be possible.

In general, components aren't responsible for making sure other components have returned sane results. If `WeightSensor` returns a value from `reservoir_weight()` without throwing an exception, then anything that calls that should be able to trust that unless it has *more* information about the context where a particular value wouldn't make sense.

Given that, I think we've actually made a mistake in splitting up this responsibility between `Adc` and the components that rely on `Adc`. Because the `read_pin` should never return a value outside of the (0, 1023) range, so we should push that check into `read_pin` and let things like `MoistureSensor` pass it through. Because from the perspective of `MoistureSensor` anything that `Adc` reports is valid.

But with something like `LightSensor` that class knows more about the pin value than `Adc` so it should be applying additional checking. Because an ADC pin can in general give a value of 5.0, but it never should if it's connected to a light sensor, so `LightSensor` should be checking for this.

So I think after this PR, we should revise the behavior of a few components so that it's like:

```python
class Adc(object):
...
  read_pin(self, pin):
      pin_value = RPi.GPIO.read(pin)
      if not (0.0 <= pin_value <= 1023.0):
         raise ValueError('whatever')

class MoistureSensor(object):
...
  def moisture(self):
    return self._adc.read_pin(adc.PIN_MOISTURE_SENSOR)

class LightSensor(object):
...
  def get_light_level(self):
        if light_level < _LIGHT_SENSOR_MIN_VALUE:
            raise ValueError('Light sensor reading out of range')
```

---

*[greenpithumb/reservoir.py, line 25 \[r3\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/23#-KJXNWvX48XxzScUX_EK:-KJXNWvYpR65vWguOyKS:-395289571) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/efffc278d69dfdadaf1f527d6386cf3280d5bdea/greenpithumb/reservoir.py#L25)):*
> ```Python
>                              water_weight_grams)
> 
>         return water_weight_grams  #1 gr water = 1 ml water
> ```

Supposably this is only at a certain temperature:

http://water.usgs.gov/edu/density.html

Should we use a multiplier of `0.99802` since that's closer to what it'll generally be indoors? If we want to go really fancy we can hook this up to `TemperatureSensor`  and do conversion as a function of temperature but that's probably overkill.

---

*[greenpithumb/weight_sensor.py, line 3 \[r3\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/23#-KJXPuUov_bgOLfwULON:-KJXPuUov_bgOLfwULOO:844858625) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/efffc278d69dfdadaf1f527d6386cf3280d5bdea/greenpithumb/weight_sensor.py#L3)):*
> ```Python
> import adc
> 
> MIN_RES_WEIGHT = 0.0
> ```

If we take out the check in `reservoir.py` this can be private again.

---

*[greenpithumb/weight_sensor.py, line 4 \[r3\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/23#-KJXPzu5w3UuT7wc-MgB:-KJXPzu5w3UuT7wc-MgC:-1333747583) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/efffc278d69dfdadaf1f527d6386cf3280d5bdea/greenpithumb/weight_sensor.py#L4)):*
> ```Python
> 
> MIN_RES_WEIGHT = 0.0
> _MAX_RES_WEIGHT = 1023.0
> ```

(optional) Kinda same note about abbreviation. I try to avoid abbreviation when I can because it generally creates ambig and conf.

---

*[greenpithumb/weight_sensor.py, line 20 \[r3\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/23#-KJXQIgXjSh0hhDBtJK9:-KJXQIgXjSh0hhDBtJKA:-2102340269) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/efffc278d69dfdadaf1f527d6386cf3280d5bdea/greenpithumb/weight_sensor.py#L20)):*
> ```Python
> 
>     def reservoir_weight(self):
>         """Returns reservoir weight in grams."""
> ```

Now that we've split it up, this can just be a generic weight sensor. It doesn't need to know that it's weight of the reservoir.

---

*[tests/test_reservoir.py, line 21 \[r3\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/23#-KJXQiBTeZYdDt06sADv:-KJXQiBTeZYdDt06sADw:-669885072) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/efffc278d69dfdadaf1f527d6386cf3280d5bdea/tests/test_reservoir.py#L21)):*
> ```Python
>         self.mock_weight_sensor.reservoir_weight.return_value = res_weight
>         water_level = res.reservoir_level()
>         self.assertAlmostEqual(water_level, res_weight - container_weight)
> ```

It's a bit of a red flag when our test code reimplements logic from the production code it's testing. If we made a mistake in production, we're probably making the same mistake in the test.

In this case, subtracting the container weight from the total weight is the most important thing `Reservoir` does, but the code is in both the production class and test class.

For test code, it's okay to introduce some redundancy for the sake of minimizing logic, so for this I'd suggest something like:

```python
self.mock_weight_sensor.reservoir_weight.return_value = 55.0
res = reservoir.Reservoir(self.mock_weight_sensor, 50)
self.assertAlmostEqual(5.0, res.reservoir_level())
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/23)*

mtlynch:
Review status: all files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[greenpithumb/reservoir.py, line 5 \[r3\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/23#-KJXOMTaJ0jm-Oz3JATW:-KJXSe3Qg0jX9194ow3V:-1218148184) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/efffc278d69dfdadaf1f527d6386cf3280d5bdea/greenpithumb/reservoir.py#L5)):*
> Wrapper probably isn't quite the right term here. Wrapper makes sense for things like the weight sensor because there actually is a physical sensor that gives out signals and we're wrapping it with a little bit of code so that there's a Python interface to represent the signals it gives off.
> 
> `Reservoir` is getting a bit more abstract because there's no "reservoir sensor" that we're calling to get this value. We're creating this concept by putting together other information. (ditto for `ReservoirMonitor`).
> 
> Suggestion:
> 
> ```python
> """Provides information about a GreenPiThumb's system's water reservoir."""
> ```
> 
> 
</details>
Whoops should be

```python
"""Provides information about a GreenPiThumb system's water reservoir."""
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/23)*

JeetShetty:
*[greenpithumb/reservoir.py, line 21 \[r3\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/23#-KJXHaFkXJaDiHakgbeH:-KKFgly1SCc11ZvukQu4:-1185091948) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/efffc278d69dfdadaf1f527d6386cf3280d5bdea/greenpithumb/reservoir.py#L21)):*
> This shouldn't be necessary. It looks like `WeightSensor` will already throw an exception in this case so this condition should never be possible.
> 
> In general, components aren't responsible for making sure other components have returned sane results. If `WeightSensor` returns a value from `reservoir_weight()` without throwing an exception, then anything that calls that should be able to trust that unless it has *more* information about the context where a particular value wouldn't make sense.
> 
> Given that, I think we've actually made a mistake in splitting up this responsibility between `Adc` and the components that rely on `Adc`. Because the `read_pin` should never return a value outside of the (0, 1023) range, so we should push that check into `read_pin` and let things like `MoistureSensor` pass it through. Because from the perspective of `MoistureSensor` anything that `Adc` reports is valid.
> 
> But with something like `LightSensor` that class knows more about the pin value than `Adc` so it should be applying additional checking. Because an ADC pin can in general give a value of 5.0, but it never should if it's connected to a light sensor, so `LightSensor` should be checking for this.
> 
> So I think after this PR, we should revise the behavior of a few components so that it's like:
> 
> ```python
> class Adc(object):
> ...
>   read_pin(self, pin):
>       pin_value = RPi.GPIO.read(pin)
>       if not (0.0 <= pin_value <= 1023.0):
>          raise ValueError('whatever')
> 
> class MoistureSensor(object):
> ...
>   def moisture(self):
>     return self._adc.read_pin(adc.PIN_MOISTURE_SENSOR)
> 
> class LightSensor(object):
> ...
>   def get_light_level(self):
>         if light_level < _LIGHT_SENSOR_MIN_VALUE:
>             raise ValueError('Light sensor reading out of range')
> ```
</details>
Good points! Agreed that we should revise ```Adc``` and the other components as necessary.  In this specific case, though, we still need to do a check because  we have the extra information of the container weight to factor in. So the weight reading could be in range, but if it's less than the weight of the container we still want to raise an exception. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/23)*

JeetShetty:
Review status: all files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[greenpithumb/reservoir.py, line 25 \[r3\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/23#-KJXNWvX48XxzScUX_EK:-KKFjsgUoi079SoVqRcj:-1297875594) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/efffc278d69dfdadaf1f527d6386cf3280d5bdea/greenpithumb/reservoir.py#L25)):*
> Supposably this is only at a certain temperature:
> 
> http://water.usgs.gov/edu/density.html
> 
> Should we use a multiplier of `0.99802` since that's closer to what it'll generally be indoors? If we want to go really fancy we can hook this up to `TemperatureSensor`  and do conversion as a function of temperature but that's probably overkill.
</details>
Oh interesting, that makes sense. I think in this instance it's fine to approximate a 1:1 relationship since precise values don't offer any practical benefit. I'm happy to change it though if you disagree. For the time-being I just updated the comment to reflect that 1 gr of water is roughly equal to 1 ml of water. 

---

*[greenpithumb/weight_sensor.py, line 3 \[r2\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/23#-KJNCbNAU_Dax7WuyDLF:-KKFgVPSJntCDN7pzUOe:-499146009) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/f2279a310268481a7e2a6592418d62538df1c9c2/greenpithumb/weight_sensor.py#L3)):*
> I think this should be a parameter to the `__init__` function since we might not have the same weight containers. This is another thing that should be in a config file eventually.
</details>
Done.

---

*[greenpithumb/weight_sensor.py, line 25 \[r2\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/23#-KJND_mv3QOw-ALZF2YH:-KKFoVcejodVJSSq_Wjp:-499146009) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/f2279a310268481a7e2a6592418d62538df1c9c2/greenpithumb/weight_sensor.py#L25)):*
> uppercase F does work, but I didn't mean to do that! Will switch to lowercase
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/23)*

mtlynch:
:lgtm:

---

Review status: 1 of 6 files reviewed at latest revision, 2 unresolved discussions.

---

*[greenpithumb/reservoir.py, line 21 \[r3\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/23#-KJXHaFkXJaDiHakgbeH:-KKakL1vjT3B2lFi0NrD:-406570654) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/efffc278d69dfdadaf1f527d6386cf3280d5bdea/greenpithumb/reservoir.py#L21)):*
Ah, yeah you're right about the container weight being the min weight that the reservoir should ever see.

In that case, `weight_sensor.MIN_WEIGHT` doesn't need to exist because it'll just be the same as  the generic check >= 0 check in `read_pin`. Then I think we should just do something like

```python
reservoir_weight_grams = self._weight_sensor.weight()
if reservoir_weight_grams < self._reservoir_weight:
  raise ReservoirUnderWeightException(
    'Unexpected weight measured for reservoir (%.1f g), as it is less than the weight of an empty reservoir (%.1f g).' % (
    reservoir_weight_grams, self._reservoir_weight))
```

I think we want a custom exception because this doesn't seem to fit the definition of a `ValueError` (we're probably stretching it in other places as well):

>Raised when a built-in operation or function receives an argument that has the right type but an inappropriate value,

https://docs.python.org/2/library/exceptions.html#exceptions.ValueError

---

*[greenpithumb/reservoir.py, line 25 \[r3\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/23#-KJXNWvX48XxzScUX_EK:-KKamyiRl9_mZavSczpz:-920948051) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/efffc278d69dfdadaf1f527d6386cf3280d5bdea/greenpithumb/reservoir.py#L25)):*
> Oh interesting, that makes sense. I think in this instance it's fine to approximate a 1:1 relationship since precise values don't offer any practical benefit. I'm happy to change it though if you disagree. For the time-being I just updated the comment to reflect that 1 gr of water is roughly equal to 1 ml of water. 
</details>
Yeah that makes sense. 1:1 approximation sounds good to me.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/23#-:-KKanMOHAmrUEYJanUzk:1417088721)*

mtlynch:
Reviewed 5 of 5 files at r4.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/23)*

JeetShetty:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[greenpithumb/reservoir.py, line 21 \[r3\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/23#-KJXHaFkXJaDiHakgbeH:-KKzA3EXbSe8pSlf-eQe:-74661253) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/efffc278d69dfdadaf1f527d6386cf3280d5bdea/greenpithumb/reservoir.py#L21)):*
> Ah, yeah you're right about the container weight being the min weight that the reservoir should ever see.
> 
> In that case, `weight_sensor.MIN_WEIGHT` doesn't need to exist because it'll just be the same as  the generic check >= 0 check in `read_pin`. Then I think we should just do something like
> 
> ```python
> reservoir_weight_grams = self._weight_sensor.weight()
> if reservoir_weight_grams < self._reservoir_weight:
>   raise ReservoirUnderWeightException(
>     'Unexpected weight measured for reservoir (%.1f g), as it is less than the weight of an empty reservoir (%.1f g).' % (
>     reservoir_weight_grams, self._reservoir_weight))
> ```
> 
> I think we want a custom exception because this doesn't seem to fit the definition of a `ValueError` (we're probably stretching it in other places as well):
> 
> >Raised when a built-in operation or function receives an argument that has the right type but an inappropriate value,
> 
> https://docs.python.org/2/library/exceptions.html#exceptions.ValueError
</details>
For the above change I guess the best thing to do is just to merge this PR as is, and then implement that change in another PR where we do the checking in ```read_pin``` and then update the rest of the classes to reflect that change?

Oh I didn't realize that was the use for ```ValueError```! What's the reason for allowing it to be manually raised in that case?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/23)*

mtlynch:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[greenpithumb/reservoir.py, line 21 \[r3\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/23#-KJXHaFkXJaDiHakgbeH:-KKza5aTf-9C003yqDIR:873706081) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/efffc278d69dfdadaf1f527d6386cf3280d5bdea/greenpithumb/reservoir.py#L21)):*
> For the above change I guess the best thing to do is just to merge this PR as is, and then implement that change in another PR where we do the checking in read_pin and then update the rest of the classes to reflect that change?

Yeah that sounds good.

>Oh I didn't realize that was the use for ValueError! What's the reason for allowing it to be manually raised in that case?

Python is super permissive, so it will *allow* us to do pretty much anything.

I can't find documentation to back me up, but I think it's accepted practice to use `ValueError` when it's basically exposing a programming error and the context is the same as when the built-in libraries raise `ValueError`.

So something like in clock.py makes sense:

```python
        if wait_time_seconds < 0.0:
            raise ValueError('Wait time cannot be negative: %f' %
                             wait_time_seconds)
```

Because it can probably only happen if we screw something up in the code and call it wrong.

Whereas checking the result of something else we're calling isn't so clearly a programming error. The code might actually be doing the right thing, but the system is in a bad state, so it makes sense to have a custom exception for those kinds of cases.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/23)*

PR #27, opened 2016-07-22
Add TemperatureStore class.

JeetShetty:
This change is [Reviewable](https://reviewable.io/reviews/jeetshetty/greenpithumb/27)


mtlynch:
Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[greenpithumb/temperature_store.py, line 3 \[r1\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/27#-KNHdMxNtpfUDkmx5bkq:-KNHdMxNtpfUDkmx5bkr:539493946) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/86722a9e2ed7c958e96517099adc290eafa3a985/greenpithumb/temperature_store.py#L3)):*
> ```Python
> import sqlite3
> ```

(nit) Seems like lots of trailing whitespace for some reason

---

*[greenpithumb/temperature_store.py, line 7 \[r1\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/27#-KNHdcfE_KONCigecusq:-KNHdcfE_KONCigecusr:-513470385) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/86722a9e2ed7c958e96517099adc290eafa3a985/greenpithumb/temperature_store.py#L7)):*
> ```Python
>     """Stores timestamp and ambient temperature readings."""
> 
>     def __init__(self, temperature_sensor, local_clock):
> ```

Storing stuff in a database is kind of hard, so I'd recommend making `TemperatureStore` as dumb as possible. I think it'll simplify things if the caller is responsible for passing `TemperatureStore` the timestamp (as a `datetime`) and a temperature (as a `float`) so that `TemperatureStore` doesn't have to interact with so many different interfaces.

---

*[greenpithumb/temperature_store.py, line 22 \[r1\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/27#-KNHewK9LsqI3QN1CglL:-KNHewKAjE0uHNmBhfgj:-1364778390) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/86722a9e2ed7c958e96517099adc290eafa3a985/greenpithumb/temperature_store.py#L22)):*
> ```Python
>         time = self._local_clock.now()
>         temperature = self._temperature_sensor.get_temperature()
>         timestamp = "%s-%s-%s %s:%s:%s" % (time.year, time.month, time.day,
> ```

I think the `sqlite3` library is clever enough to convert for us, so we can pass in a datetime and not worry about the conversion.

In general, though, if you ever want to convert a `datetime` to or from a string, check out `datetime.stftime` and `datetime.strptime` for a way to do this a little bit easier:

https://docs.python.org/2/library/datetime.html#datetime.datetime.strftime

---

*[greenpithumb/temperature_store.py, line 26 \[r1\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/27#-KNHfY83xZUlxKFaCYdd:-KNHfY83xZUlxKFaCYde:2110395005) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/86722a9e2ed7c958e96517099adc290eafa3a985/greenpithumb/temperature_store.py#L26)):*
> ```Python
> 
>         conn = sqlite3.connect("GreenPiThumb")
>         c = conn.cursor()
> ```

Suggest passing in the SQLite cursor as a parameter to `__init__`. That way, we can unit test the class by passing in a mock cursor and checking that when we call `store_temperature`, there's a corresponding call to the `execute` function of our mock with the correct SQL (kind of like what we did with `test_pump_manager.py` but a bit more sophisticated).

---

*[greenpithumb/temperature_store.py, line 27 \[r1\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/27#-KNHhDEBtWHCu3umMFWQ:-KNHhDEBtWHCu3umMFWR:83442663) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/86722a9e2ed7c958e96517099adc290eafa3a985/greenpithumb/temperature_store.py#L27)):*
> ```Python
>         conn = sqlite3.connect("GreenPiThumb")
>         c = conn.cursor()
>         c.execute("INSERT INTO temperature values(timestamp, temperature)")
> ```

We need a way of connecting the local variables to the query itself. I think we want something like this:

`execute('INSERT INTO temperature VALUES (?, ?)', (timestamp, temperature))`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/27)*

mtlynch:
Review status: all files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[greenpithumb/temperature_store.py, line 7 \[r1\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/27#-KNHdcfE_KONCigecusq:-KNHj0s-X1B18QgHSWma:1657196666) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/86722a9e2ed7c958e96517099adc290eafa3a985/greenpithumb/temperature_store.py#L7)):*
> Storing stuff in a database is kind of hard, so I'd recommend making `TemperatureStore` as dumb as possible. I think it'll simplify things if the caller is responsible for passing `TemperatureStore` the timestamp (as a `datetime`) and a temperature (as a `float`) so that `TemperatureStore` doesn't have to interact with so many different interfaces.
</details>
One other way to think about this is that relying on a temperature sensor and a clock would make it cumbersome to test because we'd have to create lots of different mocks. One of the nice things about unit testing is that in thinking about ways to make unit testing easier, we usually come up with ways to make the design simpler and more loosely coupled.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/27)*

mtlynch:
Reviewed 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[greenpithumb/temperature_store.py, line 4 \[r2\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/27#-KNJfyYQtxlGnWBi4Dh3:-KNJfyYQtxlGnWBi4Dh4:-1227046521) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/b78dfb3189419a98ec0e405018cd521fc949444a/greenpithumb/temperature_store.py#L4)):*
> ```Python
>     """Stores timestamp and ambient temperature readings."""
> 
>     def __init__(self, timestamp, temperature, cursor):
> ```

It'd be nice if we could create the `TemperatureStore` once and then keep calling `store_temperature` to add new temperatures, otherwise we have to construct the whole object every time we want to store a new value.

---

*[greenpithumb/temperature_store.py, line 11 \[r2\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/27#-KNJgA3jWIJVTcTMj6vr:-KNJgA3jWIJVTcTMj6vs:869468667) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/b78dfb3189419a98ec0e405018cd521fc949444a/greenpithumb/temperature_store.py#L11)):*
> ```Python
>                 temperature reading.
>             temperature: A float of the temperature reading in Celsius.
>             cursor: GreenPiThumb SQLite database cursor.
> ```

I'd just say `SQLite database cursor` because saying "GreenPiThumb" kind of gives the impression that we're doing something nonstandard.

---

*[tests/test_temperature_store.py, line 11 \[r2\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/27#-KNJgrSZ_buw7ow5NOqL:-KNJgrSZ_buw7ow5NOqM:-1699497329) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/b78dfb3189419a98ec0e405018cd521fc949444a/tests/test_temperature_store.py#L11)):*
> ```Python
> class TemperatureStoreTest(unittest.TestCase):
> 
>     def setUp(self):
> ```

The `setUp` can be useful if there's complicated setup logic we need to do to write tests, but in this case it's not that hard to write each test case, so suggest just putting all the logic in the test case and dropping the separate `setUp`.

---

*[tests/test_temperature_store.py, line 13 \[r2\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/27#-KNJgVo7bGH_TnyE9oJZ:-KNJgVo8SWN1nbhdwxJ2:-873170410) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/b78dfb3189419a98ec0e405018cd521fc949444a/tests/test_temperature_store.py#L13)):*
> ```Python
>     def setUp(self):
>         self.mock_timestamp = mock.Mock()
>         self.mock_temperature = mock.Mock()
> ```

We generally don't need to mock out simple types, so we can just use a real `datetime` and `float` here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/27)*

mtlynch:
Reviewed 2 of 2 files at r3.
Review status: all files reviewed at latest revision, 13 unresolved discussions, some commit checks failed.

---

*[greenpithumb/temperature_store.py, line 4 \[r3\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/27#-KNN5j_iXOwySKhh-RYo:-KNN5j_iXOwySKhh-RYp:-597231292) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/d481b20daa1c1d01b714469782d3be1008845066/greenpithumb/temperature_store.py#L4)):*
> ```Python
>     """Stores timestamp and ambient temperature readings."""
> 
>     def store_temperature(self, timestamp, temperature, cursor):
> ```

Oh, sorry I didn't mean `cursor` should be passed every time. `cursor` should be in the constructor so we can create the object once, then every time we get new temperatures, we can add them with `store_temperature`. Otherwise, you're right, if there's no permanent state, there's no need to have a class.

---

*[greenpithumb/temperature_store.py, line 13 \[r3\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/27#-KNN7l_7dKKlw1e9iyx0:-KNN7l_7dKKlw1e9iyx1:-1224289649) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/d481b20daa1c1d01b714469782d3be1008845066/greenpithumb/temperature_store.py#L13)):*
> ```Python
>             cursor: SQLite database cursor.
>         """
>         cursor.execute("INSERT INTO temperature VALUES (?, ?)",
> ```

Suggest single quotes for consistency with the other files

---

*[tests/test_temperature_store.py, line 13 \[r3\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/27#-KNN7zBz23Serpd3oMMF:-KNN7zBz23Serpd3oMMG:-1542217173) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/d481b20daa1c1d01b714469782d3be1008845066/tests/test_temperature_store.py#L13)):*
> ```Python
> 
>     def test_store_temperature_inserts_timestamp_temperature(self):
>         timestamp = datetime.datetime.now()
> ```

It probably wouldn't cause problems here, but in general unit tests should be completely deterministic, so running it multiple times should produce exactly the same results. Suggest hardcoding a specific datetime value instead of using `now()`.

---

*[tests/test_temperature_store.py, line 16 \[r3\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/27#-KNN8Mgx9tkTfKrNiQbQ:-KNN8Mgx9tkTfKrNiQbR:201412306) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/d481b20daa1c1d01b714469782d3be1008845066/tests/test_temperature_store.py#L16)):*
> ```Python
>         temperature = 21.1
>         mock_cursor = mock.Mock()
>         t_store = temperature_store.TemperatureStore()
> ```

I try to avoid ambiguous abbreviations. From context, I think `store` is clear enough here that I'd just drop the `t_`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/27)*

mtlynch:
:lgtm:

---

Reviewed 2 of 2 files at r4.
Review status: all files reviewed at latest revision, 13 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/27#-:-KNNb9qvOKwFYFqUtGx2:1417088721)*

PR #32, opened 2016-08-22
Add TemperaturePoller.

JeetShetty:
This change is [Reviewable](https://reviewable.io/reviews/jeetshetty/greenpithumb/32)


mtlynch:
Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[greenpithumb/poller.py, line 11 \[r1\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KPpOYURFI3z34VaiAuW:-KPpOYURFI3z34VaiAuX:bpoxvx3) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/4d0970d85f1101ed5fadea5bf330e22e7fe9fb3b/greenpithumb/poller.py#L11)):*
> ```Python
> 
> 		Args:
> 			local_clock: A local time zone clock interface.
> ```

Does it actually need to be local? I think any clock will do as long as it's capable of waiting.

---

*[greenpithumb/poller.py, line 22 \[r1\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KPpPULOMtNEAA4-kcdS:-KPpPULOMtNEAA4-kcdT:bshjoqa) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/4d0970d85f1101ed5fadea5bf330e22e7fe9fb3b/greenpithumb/poller.py#L22)):*
> ```Python
>     """Polls temperature data."""
> 
>     def __init__(self, temperature_sensor, temperature_store):
> ```

The `__init__` needs to also take in a `clock` and `poll_interval` and call its parent `__init__` method.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32)*

JeetShetty:
*[greenpithumb/poller.py, line 11 \[r1\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KPpOYURFI3z34VaiAuW:-KPsJuML91BTaEKYizfN:b-25tvai) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/4d0970d85f1101ed5fadea5bf330e22e7fe9fb3b/greenpithumb/poller.py#L11)):*
> Does it actually need to be local? I think any clock will do as long as it's capable of waiting.
</details>
I chose local because we're also using it for the DB timestamps

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32)*

mtlynch:
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[greenpithumb/poller.py, line 11 \[r1\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KPpOYURFI3z34VaiAuW:-KPtFfoFNdOYDPr5WcW1:b-seg46l) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/4d0970d85f1101ed5fadea5bf330e22e7fe9fb3b/greenpithumb/poller.py#L11)):*
> I chose local because we're also using it for the DB timestamps
</details>
Are we? It seems like the documentation in `db_store.py` just says timestamp and doesn't specify local or UTC (which I think is correct behavior).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32)*

JeetShetty:
*[greenpithumb/poller.py, line 11 \[r1\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KPpOYURFI3z34VaiAuW:-KPxSmIGyoPDYWX3dDsM:bxvaxic) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/4d0970d85f1101ed5fadea5bf330e22e7fe9fb3b/greenpithumb/poller.py#L11)):*
> Are we? It seems like the documentation in `db_store.py` just says timestamp and doesn't specify local or UTC (which I think is correct behavior).
</details>
hmm I'm not really sure what you're saying. ```Poller``` classes are passing the timestamp argument to the ```store``` methods, and the timestamps will make a lot more sense if they're in local time. Should I change the documentation in ```db_store``` to be more clear?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32)*

JeetShetty:
*[greenpithumb/poller.py, line 11 \[r1\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KPpOYURFI3z34VaiAuW:-KPxTnG850xmnwtaVeyR:b-4brtb2) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/4d0970d85f1101ed5fadea5bf330e22e7fe9fb3b/greenpithumb/poller.py#L11)):*
> hmm I'm not really sure what you're saying. ```Poller``` classes are passing the timestamp argument to the ```store``` methods, and the timestamps will make a lot more sense if they're in local time. Should I change the documentation in ```db_store``` to be more clear?
</details>
Oh or are you saying that the ```store``` methods are the ones that should be converting to local time?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32)*

mtlynch:
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[greenpithumb/poller.py, line 11 \[r1\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KPpOYURFI3z34VaiAuW:-KPzWN17VVQwJ-ylMF6x:b-nl5t85) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/4d0970d85f1101ed5fadea5bf330e22e7fe9fb3b/greenpithumb/poller.py#L11)):*
> Oh or are you saying that the ```store``` methods are the ones that should be converting to local time?
</details>
Oh, I was just confused because it sounded like you were saying we had to be consistent with what we did in `db_store.py` but it didn't look like we said anything about time zones at all in `db_store.py`.

I guess we can just leave it as `local_clock` if we're going to do everything in local time.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32)*

mtlynch:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[greenpithumb/poller.py, line 54 \[r2\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KQC2_K0i0dhAT894So3:-KQC2_K13IKkdg6hRGcY:b-j6mkxu) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/93647c9af06459755f6d9143c8f8b663f18c9a6e/greenpithumb/poller.py#L54)):*
> ```Python
>         t = threading.Thread(target=self._poll_forever)
>         t.setDaemon(True)
>         t.start()
> ```

We should be able to just define this function once in `SensorPollerBase` because child classes will inherit it

---

*[greenpithumb/poller.py, line 123 \[r2\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KQC4dbo4WRdzZOYNyx3:-KQC4dbpmn__U-k4twuw:b-vgnqju) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/93647c9af06459755f6d9143c8f8b663f18c9a6e/greenpithumb/poller.py#L123)):*
> ```Python
>             pumped = self._pump_manager.pump_if_needed()
>             if pumped:
>                 _WATER_PUMPED.set()
> ```

Hmm, this is tricky because it creates an implicit coupling between `MoisturePoller` and  `WateringEventPoller` and I think when we get to testing, we'll find that this makes it difficult.

Can we have `WateringEventPoller`own the pump manager?

The other thing that's kind of weird here (result of our earlier design) is that the `MoisturePoller` and `PumpManager` both independently have to poll the moisture sensor. One thing we might do instead is to either have `PumpManager` have a `SoilMoistureStore` that it can query for the latest moisture or change `pump_if_needed` to take the moisture level as a parameter (in which case `WateringEventPoller` would need to either keep a moisture sensor or a soil moisture store).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32)*

JeetShetty:
*[greenpithumb/poller.py, line 123 \[r2\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KQC4dbo4WRdzZOYNyx3:-KQLn-AjyotbVKP-BeVb:b-oqkobs) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/93647c9af06459755f6d9143c8f8b663f18c9a6e/greenpithumb/poller.py#L123)):*
Mmm good points.

>One thing we might do instead is to either have``` PumpManager``` have a ```SoilMoistureStore``` that it can query for the latest moisture

How would this work? I can't work out how to get current moisture from a ```SoilMoistureStore``` instance. 

In light of the above my current inclination is to just combine ```MoisturePoller``` and ```WateringEventPoller``` into one class, and give it a moisture sensor and pump manager as args, and then change ```pump_if_needed``` to take the moisture as an argument

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32)*

JeetShetty:
*[greenpithumb/poller.py, line 123 \[r2\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KQC4dbo4WRdzZOYNyx3:-KQLq0NbgrwLBugARCVo:bufd2ls) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/93647c9af06459755f6d9143c8f8b663f18c9a6e/greenpithumb/poller.py#L123)):*
> Mmm good points.
> 
> >One thing we might do instead is to either have``` PumpManager``` have a ```SoilMoistureStore``` that it can query for the latest moisture
> 
> How would this work? I can't work out how to get current moisture from a ```SoilMoistureStore``` instance. 
> 
> In light of the above my current inclination is to just combine ```MoisturePoller``` and ```WateringEventPoller``` into one class, and give it a moisture sensor and pump manager as args, and then change ```pump_if_needed``` to take the moisture as an argument
</details>
Oh I guess it would obviously have a moisture sensor since it's half ```MoisturePoller```!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32)*

mtlynch:
Review status: 0 of 4 files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[greenpithumb/poller.py, line 123 \[r2\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KQC4dbo4WRdzZOYNyx3:-KQOgfjKhiPQ7XlFJXpV:be5c1dv) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/93647c9af06459755f6d9143c8f8b663f18c9a6e/greenpithumb/poller.py#L123)):*
> Oh I guess it would obviously have a moisture sensor since it's half ```MoisturePoller```!
</details>
I think we're moving in the right direction in that we've gotten rid of the implicit coupling, but now there are a few "code smells" (see [here](http://c2.com/cgi/wiki?CodeSmell) and [here](https://en.wikipedia.org/wiki/Code_smell#Common_code_smells)):

* `MoistureWateringEventPoller`'s `__init__` takes six parameters, which is pretty high (sometimes okay, but usually is a hint to refactor)
 * I suspect this will give us headaches when it comes time to unit test it
* `MoistureWateringEventPoller` kind of violates [Curly's Law](https://blog.codinghorror.com/curlys-law-do-one-thing/) in that it's now hard to describe it without kind of AND'ing together two mostly disjoint things, because it polls for moisture readings to store them *and* talks to the pump manager then logs pump events.

Here are some ideas in ascending order of difficulty:

* Split up `MoisturePoller` and `WateringEventPoller` and just give each of them a moisture sensor instance
 * Maybe this is a problem because we're hitting the sensor twice as much, but maybe that's not a big deal
* Do the above, but define a "caching" moisture sensor
 * I'm thinking something kind of like this:
  
  ```python
  # Poll the real sensor if it has been >= this many seconds since the last
  # reading. Otherwise, return the last reading.
  _FRESHNESS_THRESHOLD = 10

  class CachingMoistureSensor(object):

    def __init__(self, moisture_sensor, clock):
      self._moisture_sensor = moisture_sensor
      self._clock = clock
      self._last_reading_time = datetime.min
      self._last_reading = None
      self._lock = threading.Lock()

    def moisture(self):
      self._lock.acquire()
      now = self._clock.now()
      if (now - self._last_reading_time).total_seconds >= _FRESHNESS_THRESHOLD:
        self._last_reading_time = now
        self._last_reading = self._moisture_sensor.moisture()
      self._lock.release()
      return self._last_reading
    ```
    * So then both classes share a single instance of `CachingMoistureSensor` and they can call it like it's a normal moisture sensor, but under the covers, it only calls the physical moisture sensor at most once every 10 seconds and otherwise just returns whatever it read in the last 10 seconds.
   * This is a common software pattern called the "Decorator Pattern" (not to be confused with Python Decorators). I couldn't find a good reference online ([this](https://sourcemaking.com/design_patterns/decorator) one is okay. The best is probably in [*Head First Design Patterns*](http://smile.amazon.com/dp/0596007124) (also available on Safari Books Online).
* Split up the classes, but give `WateringEventPoller` just a soil moisture store, but no sensor.
  * `SoilMoistureStore` currently just lets callers write to the store, but we can add code to let callers read from the store as well (we'll have to for the frontend anyway)
  * So then `WateringEventPoller` can just poll the database for the last recorded moisture level and pass that to the pump manager

---

*[greenpithumb/poller.py, line 80 \[r3\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KQOd_ylgkm6dFR_eM33:-KQOd_ylgkm6dFR_eM34:b-hog9kb) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/0e26d33fed6dfd8b1183c3681b02c412b45bbc1f/greenpithumb/poller.py#L80)):*
> ```Python
>                                                 humidity)
> 
>             self._local_clock.wait(self._poll_interval)
> ```

One thought since all of our `_poll_forever` functions have a similar structure is that we might want to push more functionality into the base class, so like we have a base function that's like

```python
class SensorPollerBase(object):

  def _poll_forever(self):
    while True:
      self._poll_once()
      self._local_clock.wait(self._poll_interval)

class HumidityPoller(SensorPollerBase):

  def _poll_once(self):
    humidity = self._humidity_sensor.get_humidity_level()
    self._humidity_store.store_humidity(self._local_clock.now(),
                                                humidity)
```

---

*[greenpithumb/poller.py, line 83 \[r3\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KQOf_6s16jE4gSH05X2:-KQOf_6s16jE4gSH05X3:b-cej8k) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/0e26d33fed6dfd8b1183c3681b02c412b45bbc1f/greenpithumb/poller.py#L83)):*
> ```Python
> 
> 
> class MoistureWateringEventPoller(SensorPollerBase):
> ```

Ah, gotta watch that moisture water! Suggest cutting "Moisture"?

---

*[greenpithumb/poller.py, line 105 \[r3\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KQOein-V9h5A6IqaX9u:-KQOein0cKHov0g_kEi_:b-445z9c) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/0e26d33fed6dfd8b1183c3681b02c412b45bbc1f/greenpithumb/poller.py#L105)):*
> ```Python
>         self._soil_moisture_store = soil_moisture_store
>         self._watering_event_store = watering_event_store
>         self._pump_manager = pump_manager
> ```

Suggest reordering to match the parameter order

---

*[tests/test_poller.py, line 12 \[r3\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KQOq7fw3zODXrhpeqH3:-KQOq7fw3zODXrhpeqH4:b-s8aebt) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/0e26d33fed6dfd8b1183c3681b02c412b45bbc1f/tests/test_poller.py#L12)):*
> ```Python
> 
> 
> class PollerClassesTest(unittest.TestCase):
> ```

Oooh, good first pass. You figured out that you need `side_effect` which is important.

I'll add notes below that are going to be varying degrees of cryptic so that you can still get the benefit of working through it, but let me know if you want me to be less cryptic.

---

*[tests/test_poller.py, line 22 \[r3\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KQOqnzC5y9spl8AhwlG:-KQOqnzC5y9spl8AhwlH:b-dbtkqo) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/0e26d33fed6dfd8b1183c3681b02c412b45bbc1f/tests/test_poller.py#L22)):*
> ```Python
> 
>     def temperature_poller_thread(self):
>         while True:
> ```

We probably don't want to loop forever in our test. Is there a way we can verify that just one poll event happened?

---

*[tests/test_poller.py, line 23 \[r3\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KQOr0ckK4NWSYNQFD5J:-KQOr0ckK4NWSYNQFD5K:b1ir2zl) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/0e26d33fed6dfd8b1183c3681b02c412b45bbc1f/tests/test_poller.py#L23)):*
> ```Python
>     def temperature_poller_thread(self):
>         while True:
>             lock = threading.Lock()
> ```

Locks let us coordinate actions across threads. If I'm reading this correctly, only one thread ever executes this code, so the lock doesn't do anything here. When we use locks, we want to figure out how multiple threads are going to share that same lock.

---

*[tests/test_poller.py, line 39 \[r3\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KQOqYYTAAWjFuwuhGmC:-KQOqYYUDaZhLnh_QtV6:bhrz6fs) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/0e26d33fed6dfd8b1183c3681b02c412b45bbc1f/tests/test_poller.py#L39)):*
> ```Python
>         self.mock_local_clock.now.return_value = now
>         self.mock_temperature_sensor.get_temperature.return_value = 21.0
>         self.mock_local_clock.wait.side_effect = self.wait
> ```

This is a good spot to use a lambda function. I think we can do:

```python
self.mock_local_clock.wait.side_effect = lambda x: time.sleep(x)
```

Except I'll say it's not really important that `wait` actually causes the thread to go to sleep as much as it's important that the main test thread can verify `wait` was called.

---

*[tests/test_poller.py, line 45 \[r3\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KQOryXY2Abqr6s6Ese9:-KQOryXY2Abqr6s6EseA:bldyijw) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/0e26d33fed6dfd8b1183c3681b02c412b45bbc1f/tests/test_poller.py#L45)):*
> ```Python
>         t.start()
> 
>         time.sleep(10)
> ```

We generally don't want to sleep in the main testing thread because it means that running this one unit test will take > 10 seconds (when most of our tests run in milliseconds).

We *want* to sleep because we need to give the background thread time to get to the sensor polling code. So it would be great if instead of sleeping for an arbitrary amount of time, we could wait until the background thread signalled to us that it had finished a polling interval. Can we use locks to let the background thread signal to the main thread that this happened?

---

*[tests/test_pump_manager.py, line 15 \[r3\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KQOgCr9mOsoxBv366Jq:-KQOgCr9mOsoxBv366Jr:bdi4x79) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/0e26d33fed6dfd8b1183c3681b02c412b45bbc1f/tests/test_pump_manager.py#L15)):*
> ```Python
>         self.mock_pump_scheduler = mock.Mock()
>         self.manager = pump_manager.PumpManager(self.mock_pump,
>                                                 self.mock_pump_scheduler)
> ```

The fact that this simplified our tests is a good sign that this is a good design adjustment

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32)*

mtlynch:
Reviewed 1 of 4 files at r3, 5 of 5 files at r4.
Review status: all files reviewed at latest revision, 22 unresolved discussions, some commit checks failed.

---

*[greenpithumb/db_store.py, line 27 \[r4\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KQmx-nPxX8wSD8NxfLq:-KQmx-nPxX8wSD8NxfLr:bpifnpi) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/ff2822266ac1e66503762f1ff6e7ea39bc69a670/greenpithumb/db_store.py#L27)):*
> ```Python
>                              (timestamp, soil_moisture))
> 
>     def recent_soil_moisture(self):
> ```

Suggest `get_latest_soil_moisture` so that it has similar naming to `store_soil_moisture`

---

*[greenpithumb/db_store.py, line 36 \[r4\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KQn-8J3YLmK0uoyKVMF:-KQn-8J3YLmK0uoyKVMG:bg4jfn1) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/ff2822266ac1e66503762f1ff6e7ea39bc69a670/greenpithumb/db_store.py#L36)):*
> ```Python
>         return self._cursor.fetchall()[0][0]
> ```

We don't need to change anything for this PR, but a change like this probably makes sense as an independent PR. This PR is still small enough that both changes together are only ~300 LOC changing, but it's good to think about what changes are separate and can go in independently. Smaller, tightly scoped PRs are faster to review and they make it easier to read the change history later on.

---

*[greenpithumb/poller.py, line 89 \[r4\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KQmxYpCtnxkAsS0EKm_:-KQmxYpDsJySUMfc49OP:b-9y3uyu) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/ff2822266ac1e66503762f1ff6e7ea39bc69a670/greenpithumb/poller.py#L89)):*
> ```Python
>         """Creates a MoisturePoller object.
> 
> 		Args:
> ```

tabs (a few times in this file)

---

*[greenpithumb/poller.py, line 160 \[r4\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KQn0WxmtdmBgdZZd8ag:-KQn0WxmtdmBgdZZd8ah:b-3ync22) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/ff2822266ac1e66503762f1ff6e7ea39bc69a670/greenpithumb/poller.py#L160)):*
> ```Python
> 
> class WateringEventPoller(SensorPollerBase):
>     """Polls watering event data."""
> ```

Suggest elaborating a bit more in these class docstrings. `WateringEventPoller` is kind of a special case, but from the docstring it sounds like it's like all the others.

---

*[greenpithumb/poller.py, line 181 \[r4\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KQn-zshp0QXlahn3mg0:-KQn-zshp0QXlahn3mg1:bf1mjpl) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/ff2822266ac1e66503762f1ff6e7ea39bc69a670/greenpithumb/poller.py#L181)):*
> ```Python
> 
>     def _poll_once(self):
>         """Calls pump_if_needed, polls watering event data if pump is run."""
> ```

Suggest describing this without using "pump_if_needed." If someone's reading that and doesn't know what `pump_manager` looks like, it'd be difficult for them to get a sense of what this function does without reading the function's code.

Also suggest rephrasing the description of what it does and probably adding an explanation beyond the one-line summary. "polls watering event data" is kind of vague to the reader and seems to contradict what this function actually does, since what we're actually polling is updates to moisture information and we're *recording* when the pump runs.

---

*[greenpithumb/poller.py, line 186 \[r4\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KQn1VwiUtcBslgZOBUN:-KQn1Vwj4vvYyV1Ur4Q5:b-kdy4wu) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/ff2822266ac1e66503762f1ff6e7ea39bc69a670/greenpithumb/poller.py#L186)):*
> ```Python
>         if pumped:
>             self._watering_event_store.store_water_pumped(
>                 self._local_clock.now(), pump_manager.DEFAULT_PUMP_AMOUNT)
> ```

Looking at this now, the semantics are a little weird. It feels like a violation of encapsulation that clients of `pump_manager` have to know that when the pump runs, the amount that was pumped is equal to this constant. What if `pump_if_needed` actually returned a float of how much was pumped, so 0.0 means nothing was pumped?

---

*[tests/test_db_store.py, line 13 \[r4\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KQmyCu8JsepM36GD_B1:-KQmyCu8JsepM36GD_B2:b-z3c2a3) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/ff2822266ac1e66503762f1ff6e7ea39bc69a670/tests/test_db_store.py#L13)):*
> ```Python
> 
>     def setUp(self):
>         self.mock_cursor = mock.Mock()
> ```

(optional) I've recently been leaning toward more redundancy in tests if it lets each test be completely self-contained in the unit test function. I've realized there's a lot of value in being able to see every single thing about a test without having to jump around in the file.

For example, if I'm reading the original version of `test_soil_moisture`, I can read the function top to bottom and know everything that's happening. In the revised version, when I see `self.mock_cursor`, I have to jump up to the `setUp` function to see how we've initialized `self.mock_cursor`.

---

*[tests/test_db_store.py, line 31 \[r4\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KQmzsnOE249EW-jWmks:-KQmzsnOE249EW-jWmkt:b-e3qctl) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/ff2822266ac1e66503762f1ff6e7ea39bc69a670/tests/test_db_store.py#L31)):*
> ```Python
>         self.assertTrue(self.mock_cursor.execute.called)
>         self.assertEqual(300, moisture)
> ```

We should add a test for when the database has no rows, so the return value is just `[]`.

---

*[tests/test_poller.py, line 24 \[r4\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KQn3BfyX3FrF57TMUu3:-KQn3BfzP3fe0pJ0gL4H:bejamzu) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/ff2822266ac1e66503762f1ff6e7ea39bc69a670/tests/test_poller.py#L24)):*
> ```Python
>             self.mock_temperature_sensor, self.mock_temperature_store)
>         now = datetime.datetime(2016, 7, 23, 10, 51, 9, 928000)
>         self.mock_local_clock.now.return_value = now
> ```

Suggest reducing to just 

```python
self.mock_local_clock.now.return_value = datetime.datetime(2016, 7, 23, 10, 51, 9, 928000)
```

---

*[tests/test_poller.py, line 31 \[r4\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KQn3QtVcuYifRb33JAb:-KQn3QtVcuYifRb33JAc:b-fkodam) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/ff2822266ac1e66503762f1ff6e7ea39bc69a670/tests/test_poller.py#L31)):*
> ```Python
>         while True:
>             if self.mock_temperature_store.store_temperature.called:
>                 break
> ```

Oh, this is pretty clever! I think this is better than what I was thinking.

My idea was something like this:

```python
# Setup stuff you already have (elided)
self.mock_temperature_sensor = mock.Mock()
...
self.mock_temperature_sensor.get_temperature.return_value = 21.0

# Create a lock to signal between threads.
lock = threading.Lock()
# Set up mock_local_clock.wait so that it releases the lock when it's
# called. We're taking advantage of Python closures here:
# http://www.shutupandship.com/2012/01/python-closures-explained.html
# We're using the variable name _ because wait takes a parameter, but
# we'll ignore it, and _ is the name for an ignored parameter:
# https://google.github.io/styleguide/pyguide.html?showone=Lint#Lint
self.mock_local_clock.wait.side_effect = lambda _: lock.release()

# Acquire the lock so that we can't acquire again until it's released.
lock.acquire()
# Start the temperature poller, which should release the lock the
# first time that clock.wait() is called.
temperature_poller.start_polling_async()
# Try to acquire the lock. We can't acquire it until lock.release gets
# called from the background thread, so once we get past this call,
# we know the background thread has completed one poll interval.
lock.acquire()

self.mock_temperature_store.store_temperature.assert_called_once_with(
            datetime.datetime(2016, 7, 23, 10, 51, 9, 928000), 21.0)
self.mock_local_clock.wait.assert_called_once_with(self.poll_interval)
```

But I think your solution is better. The only advantage to the locks approach is that it's a little more deterministic in that locks are thread-safe, whereas mock objects are not. So with mock objects, `wait` might actually be called in the background thread, but in the main thread `wait.called` is still `False` because mock isn't thread-safe and so isn't doing anything special to coordinate reads and writes of the object. In contrast, locks are thread-safe so there can't be issues like where one thread calls lock.release, but the other thread still thinks the lock is being held.

But in this case, not being thread-safe doesn't really matter because the main thread will *eventually* see `called` as `True`. The delay is probably just one loop iteration, which is like microseconds, so it doesn't matter.

In either case, we have the problem of an infinite loop if the thing we're expecting to get called never gets called. We can solve that by putting in a timeout (like loop for max 2 seconds) but I'd say we can put off adding in that logic until we actually run into a case of that.

---

*[tests/test_poller.py, line 33 \[r4\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KQnDMI-OEtGITtzRxqG:-KQnDMI-OEtGITtzRxqH:b8vw009) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/ff2822266ac1e66503762f1ff6e7ea39bc69a670/tests/test_poller.py#L33)):*
> ```Python
>                 break
>         self.mock_temperature_store.store_temperature.assert_called_once_with(
>             datetime.datetime(2016, 7, 23, 10, 51, 9, 928000), 21.0)
> ```

The changes I'd suggest are:

* waiting based on `clock.wait.called` instead of `store_temperature.called` because we're making assertions about a full polling interval and `clock.wait` indicates a complete polling interval
* take out the call to `time.sleep` in the `wait.side_effect` because I don't think the thread actually needs to sleep
* add in `self.mock_local_clock.wait.assert_called_once_with(self.poll_interval)`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32)*

JeetShetty:
*[greenpithumb/poller.py, line 181 \[r4\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KQn-zshp0QXlahn3mg0:-KR3X_Tpq2gIZESp8HOS:b-3m6fjy) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/ff2822266ac1e66503762f1ff6e7ea39bc69a670/greenpithumb/poller.py#L181)):*
Ah I'm glad you brought up the docstrings! I find them difficult and mostly feel like I'm just writing something down so that it's not empty without actually thinking they're any good.

>Also suggest rephrasing the description of what it does and probably adding an explanation beyond the one-line summary. "polls watering event data" is kind of vague to the reader and seems to contradict what this function actually does, since what we're actually polling is updates to moisture information and we're recording when the pump runs.

Agreed! But then, what do we do about the class name? On the one hand I think keeping the class names consistent makes it easier on someone reading the code, bc once they understand what one does then it's clear that all the other classes with that naming are achieving a similar function. But if we do that, then the class name isn't really reflecting what it does. Maybe just by having them all in the same file it's clear they're doing similar things, so we can name them whatever we want? Also if ```WateringEventPoller``` did have a different name, what would it be? Mentioning polling soil moisture seems weird since there is already a ```MoisturePoller```, and I'm also not sure how to keep things succinct while still completely describing the function. 

I also get confused on redundancy. Because the ```Poller``` classes are actually polling and storing, but there are already store classes that do the actual storing.  So if you put in a docstring that ```_poll_once``` is storing data, does it become confusing that there are 2 functions that both claim to be storing the same data? I guess I'm just generally wondering how to distinguish between a method calling another method to do something, vs it actually doing it within the method. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32)*

JeetShetty:
*[greenpithumb/poller.py, line 186 \[r4\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KQn1VwiUtcBslgZOBUN:-KR3_t9W8evRGWpMg92y:b-ti962x) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/ff2822266ac1e66503762f1ff6e7ea39bc69a670/greenpithumb/poller.py#L186)):*
> Looking at this now, the semantics are a little weird. It feels like a violation of encapsulation that clients of `pump_manager` have to know that when the pump runs, the amount that was pumped is equal to this constant. What if `pump_if_needed` actually returned a float of how much was pumped, so 0.0 means nothing was pumped?
</details>
Good point, I like this!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32)*

JeetShetty:
*[greenpithumb/poller.py, line 181 \[r4\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KQn-zshp0QXlahn3mg0:-KR3ao0q2M6NX3etyvPk:b-rfx8ao) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/ff2822266ac1e66503762f1ff6e7ea39bc69a670/greenpithumb/poller.py#L181)):*
>Maybe just by having them all in the same file it's clear they're doing similar things, so we can name them whatever we want?

I mean since they're all children of the same parent class!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32)*

JeetShetty:
*[tests/test_poller.py, line 31 \[r4\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KQn3QtVcuYifRb33JAb:-KR3er-5x0MGQVHshCdh:b-ybk3yb) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/ff2822266ac1e66503762f1ff6e7ea39bc69a670/tests/test_poller.py#L31)):*
> Oh, this is pretty clever! I think this is better than what I was thinking.
> 
> My idea was something like this:
> 
> ```python
> # Setup stuff you already have (elided)
> self.mock_temperature_sensor = mock.Mock()
> ...
> self.mock_temperature_sensor.get_temperature.return_value = 21.0
> 
> # Create a lock to signal between threads.
> lock = threading.Lock()
> # Set up mock_local_clock.wait so that it releases the lock when it's
> # called. We're taking advantage of Python closures here:
> # http://www.shutupandship.com/2012/01/python-closures-explained.html
> # We're using the variable name _ because wait takes a parameter, but
> # we'll ignore it, and _ is the name for an ignored parameter:
> # https://google.github.io/styleguide/pyguide.html?showone=Lint#Lint
> self.mock_local_clock.wait.side_effect = lambda _: lock.release()
> 
> # Acquire the lock so that we can't acquire again until it's released.
> lock.acquire()
> # Start the temperature poller, which should release the lock the
> # first time that clock.wait() is called.
> temperature_poller.start_polling_async()
> # Try to acquire the lock. We can't acquire it until lock.release gets
> # called from the background thread, so once we get past this call,
> # we know the background thread has completed one poll interval.
> lock.acquire()
> 
> self.mock_temperature_store.store_temperature.assert_called_once_with(
>             datetime.datetime(2016, 7, 23, 10, 51, 9, 928000), 21.0)
> self.mock_local_clock.wait.assert_called_once_with(self.poll_interval)
> ```
> 
> But I think your solution is better. The only advantage to the locks approach is that it's a little more deterministic in that locks are thread-safe, whereas mock objects are not. So with mock objects, `wait` might actually be called in the background thread, but in the main thread `wait.called` is still `False` because mock isn't thread-safe and so isn't doing anything special to coordinate reads and writes of the object. In contrast, locks are thread-safe so there can't be issues like where one thread calls lock.release, but the other thread still thinks the lock is being held.
> 
> But in this case, not being thread-safe doesn't really matter because the main thread will *eventually* see `called` as `True`. The delay is probably just one loop iteration, which is like microseconds, so it doesn't matter.
> 
> In either case, we have the problem of an infinite loop if the thing we're expecting to get called never gets called. We can solve that by putting in a timeout (like loop for max 2 seconds) but I'd say we can put off adding in that logic until we actually run into a case of that.
</details>
Ah OK, yeah I thought of your idea but I was skeptical of adding functionality as a side effect that doesn't exist in the production code. I feel like another problem with both approaches is that there's no way to assure that the tests are run before a second call to ```store_temperature```. I thought the ```time.sleep()``` side effect was a happy little coincidence that acts as insurance against that, which is why I left it in. Are we fine just assuming that the tests will run before a second call to ```store_temperature``` unless we run into problems? 

If we are concerned with the timing, I think your solution is better than mine, since ```store_temperature``` is called 2 times at most using your test which we can probably adjust the test to account for, where as with mine it's unbounded.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32)*

JeetShetty:
*[tests/test_poller.py, line 31 \[r4\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KQn3QtVcuYifRb33JAb:-KR3wKo6Bx0HEpNjFldW:b-6pke2r) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/ff2822266ac1e66503762f1ff6e7ea39bc69a670/tests/test_poller.py#L31)):*
>If we are concerned with the timing, I think your solution is better than mine, since store_temperature is called 2 times at most using your test which we can probably adjust the test to account for, where as with mine it's unbounded.

Oh hmm I thought that ```lock.release()``` would just wait until the lock was acquired elsewhere, but now I see that it raises a ```ThreadError```, so maybe that's not so useful if 2 poll intervals complete before the tests are run. 

Also I need to reread about closures bc I didn't get it the first time around, so if that answers any of my questions then just ignore me until I figure that out!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32)*

mtlynch:
Review status: 0 of 6 files reviewed at latest revision, 23 unresolved discussions, some commit checks failed.

---

*[greenpithumb/db_store.py, line 36 \[r5\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KR62QCHqqcIEzb-_fWx:-KR62QCHqqcIEzb-_fWy:b-ie3st9) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/fe47769a1736876f7bfe11c409a5673eee2c1cf0/greenpithumb/db_store.py#L36)):*
> ```Python
>         results = self._cursor.fetchall()
>         if results == []:
>             return 1023  #return max value to ensure pump doesn't run
> ```

This has a few problems:

* It assumes the caller's intent beyond what the function is defined to do. 
 * Could be problematic when we hook up the web page front end and we want to have something that displays the current moisture temperature
* It binds tightly with soil moisture sensor and creates a new magic number that represents 
 * Could be problematic if we decide to change the temperature sensor's units and scale to some other range (like 0-100,000).

I think the right solution here is to modify the behavior so that it makes sense in general rather than specifically in the context of `pump_manager`. Can it return `None` when there's no stored history? Alternatively, it could raise a custom exception, but `None` is simpler and seems like reasonable behavior.

---

*[greenpithumb/poller.py, line 181 \[r4\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KQn-zshp0QXlahn3mg0:-KR64zCTplttFA-ahJ84:b-cn8uy6) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/ff2822266ac1e66503762f1ff6e7ea39bc69a670/greenpithumb/poller.py#L181)):*
> Agreed! But then, what do we do about the class name? On the one hand I think keeping the class names consistent makes it easier on someone reading the code, bc once they understand what one does then it's clear that all the other classes with that naming are achieving a similar function. But if we do that, then the class name isn't really reflecting what it does. Maybe just by having them all in the same file it's clear they're doing similar things, so we can name them whatever we want? Also if WateringEventPoller did have a different name, what would it be?

I think `WateringEventPoller` is a pretty good name. The downside is that with the other classes, it kind of implies that it's passively observing something, whereas this is actually polling something, causing an action, then recording that. But I think it's different enough that the reader should recognize that it breaks the pattern of pure sensors (e.g. `MoisturePoller`, `TemperaturePoller`). My only other thought would be something like `WaterPumpPoller` but it kind of has the same problem.

One thing I try to think about when I'm naming things according to a pattern is that the reader should be able to assume everything that matches a pattern has some commonalities, but the non-pattern part of each thing should convey how it varies within the pattern. So like `TemperaturePoller` and `MoisturePoller` fit the pattern of [environmental characteristic]+Poller and they're pretty much exactly the same except one is temperature and one is moisture. `WateringEventPoller` doesn't match this pattern, but matches the more general pattern of [thing]+poller, and it does seem to honor the pattern of repeatedly performing polling and reacting to some information on an interval.

>I also get confused on redundancy. Because the Poller classes are actually polling and storing, but there are already store classes that do the actual storing. So if you put in a docstring that _poll_once is storing data, does it become confusing that there are 2 functions that both claim to be storing the same data? 

I don't think this is an issue. We could maybe be more specific and say that it's "recording the event in the watering event store." But having two things that both claim to store the same information isn't really a problem because there are always going to be layers of abstraction on anything we do. We have a function that stores something in a SQLite database, but if we kept reading into the SQLite source, we'd find something that stores the database on the filesystem, then the OS has a function to store the file on the physical disk and so on. [It's abstractions all the way down](https://en.wikipedia.org/wiki/Turtles_all_the_way_down)!

---

*[tests/test_db_store.py, line 37 \[r5\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KR68Mo9QgHtWVJA62HP:-KR68Mo9QgHtWVJA62HQ:b2kfdoy) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/fe47769a1736876f7bfe11c409a5673eee2c1cf0/tests/test_db_store.py#L37)):*
> ```Python
>         moisture = store.get_latest_soil_moisture()
>         self.assertTrue(mock_cursor.execute.called)
>         self.assertEqual(1023, moisture)
> ```

To further pick on the `1023`, the reader should usually be able to understand what a function does just by reading its unit tests, so it's a red flag that the 1023 would be really mystifying to a reader who didn't read the function implementation.

---

*[tests/test_poller.py, line 31 \[r4\]](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KQn3QtVcuYifRb33JAb:-KR69Xk4CbzQZsHnSASS:bunx2qh) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/ff2822266ac1e66503762f1ff6e7ea39bc69a670/tests/test_poller.py#L31)):*
> Ah OK, yeah I thought of your idea but I was skeptical of adding functionality as a side effect that doesn't exist in the production code. 

Oh, you don't have to worry about that. Mocks can do whatever they want. As long as the behavior they present to the system under test is something that the real object could actually do (e.g. the mock shouldn't return strings if the real object actually returns floats), we can go nuts.

>Oh hmm I thought that lock.release() would just wait until the lock was acquired elsewhere, but now I see that it raises a ThreadError, so maybe that's not so useful if 2 poll intervals complete before the tests are run.

Oh, maybe the better solution here would be `threading.Event`
https://docs.python.org/2/library/threading.html#event-objects

Then the lambda function could set() the event and in main test thread, instead of waiting for the lock, we just call like `clock_wait_event.wait()`. It also has the nice side effect that it's easy for us to time out and fail if the event doesn't fire in a reasonable timeframe (like 3 seconds).

>If we are concerned with the timing, I think your solution is better than mine, since store_temperature is called 2 times at most using your test which we can probably adjust the test to account for, where as with mine it's unbounded.

I think it's a good thought to try to make sure we're only asserting after a single poll interval, but I also think it might be more trouble than it's worth to try to ensure that's the case. For any test, we kind of have to weight the difficulty of writing the test against the value it gets us. I think for this one, I'm satisfied with a kind of rough test that makes sure the background thread gets at least one of the calls right. If we end up with bugs in this function, maybe it makes sense to add more rigorous tests to make sure that each poll interval corresponds to exactly one set of calls. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32)*

JeetShetty:
*[greenpithumb/db_store.py, line 36 at r5](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KR62QCHqqcIEzb-_fWx:-KR_4M-cBj2yDXR6Bb60:b-komjdy) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/fe47769a1736876f7bfe11c409a5673eee2c1cf0/greenpithumb/db_store.py#L36)):*
> This has a few problems:
> 
> * It assumes the caller's intent beyond what the function is defined to do. 
>  * Could be problematic when we hook up the web page front end and we want to have something that displays the current moisture temperature
> * It binds tightly with soil moisture sensor and creates a new magic number that represents 
>  * Could be problematic if we decide to change the temperature sensor's units and scale to some other range (like 0-100,000).
> 
> I think the right solution here is to modify the behavior so that it makes sense in general rather than specifically in the context of `pump_manager`. Can it return `None` when there's no stored history? Alternatively, it could raise a custom exception, but `None` is simpler and seems like reasonable behavior.
</details>
Ah yeah good points. This was sloppy by me.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32)*

JeetShetty:
*[tests/test_poller.py, line 31 at r4](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KQn3QtVcuYifRb33JAb:-KR_KRuInvufpfhph4Hm:b-1hwun4) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/ff2822266ac1e66503762f1ff6e7ea39bc69a670/tests/test_poller.py#L31)):*
> > Ah OK, yeah I thought of your idea but I was skeptical of adding functionality as a side effect that doesn't exist in the production code. 
> 
> Oh, you don't have to worry about that. Mocks can do whatever they want. As long as the behavior they present to the system under test is something that the real object could actually do (e.g. the mock shouldn't return strings if the real object actually returns floats), we can go nuts.
> 
> >Oh hmm I thought that lock.release() would just wait until the lock was acquired elsewhere, but now I see that it raises a ThreadError, so maybe that's not so useful if 2 poll intervals complete before the tests are run.
> 
> Oh, maybe the better solution here would be `threading.Event`
> https://docs.python.org/2/library/threading.html#event-objects
> 
> Then the lambda function could set() the event and in main test thread, instead of waiting for the lock, we just call like `clock_wait_event.wait()`. It also has the nice side effect that it's easy for us to time out and fail if the event doesn't fire in a reasonable timeframe (like 3 seconds).
> 
> >If we are concerned with the timing, I think your solution is better than mine, since store_temperature is called 2 times at most using your test which we can probably adjust the test to account for, where as with mine it's unbounded.
> 
> I think it's a good thought to try to make sure we're only asserting after a single poll interval, but I also think it might be more trouble than it's worth to try to ensure that's the case. For any test, we kind of have to weight the difficulty of writing the test against the value it gets us. I think for this one, I'm satisfied with a kind of rough test that makes sure the background thread gets at least one of the calls right. If we end up with bugs in this function, maybe it makes sense to add more rigorous tests to make sure that each poll interval corresponds to exactly one set of calls. 
</details>
Ah yeah I like using ```threading.Event```. I tried putting a timeout in the ```event.wait()``` call, but the tests fail when I do that(```store.store``` is called twice), so I left the timeout out for now. Not sure if I'm misunderstanding what the timeout is supposed to do. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32)*

mtlynch:
Review status: 0 of 6 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[greenpithumb/pump_manager.py, line 32 at r6](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KRfIf4UmuqpBFCWr7gx:-KRfIf4UmuqpBFCWr7gy:bgazdj9) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/20529233ff934b91c60b5b98eeddd96eda4577ea/greenpithumb/pump_manager.py#L32)):*
> ```Python
>         the defined moisture threshold. Returns a float of the amount of water
>         pumped, in ml.
>         """
> ```

We should add an `Args` section to document the `moisture` arg.

---

*[greenpithumb/pump_manager.py, line 34 at r6](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KRfInunDQxp6XlJ_vQ2:-KRfInunDQxp6XlJ_vQ3:bza9bkr) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/20529233ff934b91c60b5b98eeddd96eda4577ea/greenpithumb/pump_manager.py#L34)):*
> ```Python
>         """
>         if self._pump_scheduler.is_running_pump_allowed():
>             if (moisture != None) and (moisture < SOIL_MOISTURE_THRESHOLD):
> ```

I feel like this is unintuitive semantics for `PumpManager`. I think it should expect its callers to pass sane values and `None` as a moisture value isn't really a sane moisture value.

Can we make the change in `WateringEventPoller`? Because that's really who's responsible for checking for `None` in the result from the moisture sensor.

---

*[greenpithumb/pump_manager.py, line 36 at r6](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KRfNnxdbYqAFq--BpED:-KRfNnxdbYqAFq--BpEE:b-1l8ybf) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/20529233ff934b91c60b5b98eeddd96eda4577ea/greenpithumb/pump_manager.py#L36)):*
> ```Python
>             if (moisture != None) and (moisture < SOIL_MOISTURE_THRESHOLD):
>                 self._pump.pump_water(DEFAULT_PUMP_AMOUNT)
>                 return DEFAULT_PUMP_AMOUNT
> ```

Suggest doing something like

```python
pump_amount = DEFAULT_PUMP_AMOUNT
self._pump.pump_water(pump_amount)
return pump_amount
```

That way it's more explicit that the return value is what's pumped. If they're both returning the constant, it could be interpreted as being independent values that happen to be the same.

---

*[tests/test_db_store.py, line 37 at r6](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KRfHata1xn8z3Oi0l2V:-KRfHatbLkcykfrs1opA:b-dq5vf3) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/20529233ff934b91c60b5b98eeddd96eda4577ea/tests/test_db_store.py#L37)):*
> ```Python
>         moisture = store.get_latest_soil_moisture()
>         self.assertTrue(mock_cursor.execute.called)
>         self.assertEqual(None, moisture)
> ```

We can use `self.assertIsNone(moisture)` here

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32)*

mtlynch:
Reviewed 1 of 5 files at r6, 5 of 5 files at r7.

---

*[greenpithumb/poller.py, line 188 at r7](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KRkGx-zWzP-ASif_OaH:-KRkGx-zWzP-ASif_OaI:b-xyqhn8) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/d426d9618ca15f9068bd92af4b0a5ad0706124d1/greenpithumb/poller.py#L188)):*
According to the style guide, should be `if moisture is not None:`

>Never use == or != to compare singletons like None. Use is or is not.

https://google.github.io/styleguide/pyguide.html?showone=True/False_evaluations#True/False_evaluations

---

*[greenpithumb/poller.py, line 188 at r7](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KRkHAZk1W-Fb8mxSRs3:-KRkHAZk1W-Fb8mxSRs4:b-dawjj8) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/d426d9618ca15f9068bd92af4b0a5ad0706124d1/greenpithumb/poller.py#L188)):*
> ```Python
>         """
>         moisture = self._soil_moisture_store.get_latest_soil_moisture()
>         if moisture != None:
> ```

(optional) For stuff like this, I try to save myself nesting, so another way to write this is

```python
if moisture is None:
  return

ml_pumped = self._pump_manager.pump_if_needed(moisture)
if ml_pumped <= 0:
  return


self._watering_event_store.store_water_pumped(
  self._local_clock.now(), ml_pumped)
```

But it's a matter of taste. Some people find it more readable when it's nested like that.

---

*[tests/test_poller.py, line 31 at r4](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KQn3QtVcuYifRb33JAb:-KRkI6qvjC8VkY47AR-D:bl4rk2v) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/ff2822266ac1e66503762f1ff6e7ea39bc69a670/tests/test_poller.py#L31)):*
> Ah yeah I like using ```threading.Event```. I tried putting a timeout in the ```event.wait()``` call, but the tests fail when I do that(```store.store``` is called twice), so I left the timeout out for now. Not sure if I'm misunderstanding what the timeout is supposed to do. 
</details>
Do the tests still fail with a timeout now? The timeout should do what you expect, which in this case is nothing, unless we break the code so that `clock.wait` is never called. But if we break it and there's not timeout, we'll hang forever. If there's a timeout, the test will fail after (whatever) seconds.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32)*

JeetShetty:
*[greenpithumb/poller.py, line 188 at r7](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KRkHAZk1W-Fb8mxSRs3:-KRno3Zct6gTuze3nppO:b-1hvpxg) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/d426d9618ca15f9068bd92af4b0a5ad0706124d1/greenpithumb/poller.py#L188)):*
> (optional) For stuff like this, I try to save myself nesting, so another way to write this is
> 
> ```python
> if moisture is None:
>   return
> 
> ml_pumped = self._pump_manager.pump_if_needed(moisture)
> if ml_pumped <= 0:
>   return
> 
> 
> self._watering_event_store.store_water_pumped(
>   self._local_clock.now(), ml_pumped)
> ```
> 
> But it's a matter of taste. Some people find it more readable when it's nested like that.
</details>
Interesting. I left it as is for now because I find the nesting more intuitive/readable. Why do you prefer the other way?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32)*

mtlynch:
Reviewed 2 of 2 files at r8.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[greenpithumb/poller.py, line 188 at r7](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KRkHAZk1W-Fb8mxSRs3:-KRpGw-iWIOKg62rKaii:bejbimn) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/d426d9618ca15f9068bd92af4b0a5ad0706124d1/greenpithumb/poller.py#L188)):*
> Interesting. I left it as is for now because I find the nesting more intuitive/readable. Why do you prefer the other way?
</details>
I think I just prefer reading with less nesting because I feel like every time I go down a layer of nesting, I have to keep in my head the extra state that's implied by that nested scope. Also there's the kind of silly reason that if we're sticking to 80 column lines, I like styles don't add indents and limit what I can prettily fit on a single line.

---

*[tests/test_db_store.py, line 28 at r8](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KRpKGEqsK__2_IXPHa_:-KRpKGErSGzDU2iWdiJX:b-jpm34x) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/b17c456fad2ca1a6614a70a630b8fa277597fe78/tests/test_db_store.py#L28)):*
> ```Python
>         mock_cursor.fetchall.return_value = [(300,)]
>         moisture = store.get_latest_soil_moisture()
>         self.assertTrue(mock_cursor.execute.called)
> ```

I think we should cut this line. We want to minimize the amount we care about implementation details so that our tests stay flexible. `get_latest_soil_moisture`'s job is to retrieve the right value for us (as opposed to `store_soil_moisture` whose job is to cause an effect that's not visible in the return value). If it gets the right answer, that should be enough.

Also it lets us simplify to the more readable

```python
mock_cursor.fetchall.return_value = [(300,)]
self.assertEqual(300, store.get_latest_soil_moisture())
```

(ditto for next test)

---

*[tests/test_poller.py, line 15 at r8](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KRpN3oXiZPW4Vj1GESm:-KRpN3oXiZPW4Vj1GESn:b-xwavwt) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/b17c456fad2ca1a6614a70a630b8fa277597fe78/tests/test_poller.py#L15)):*
> ```Python
> 
>     def test_temperature_poller(self):
>         clock_wait_event = threading.Event()
> ```

Hmm, we're ending up with a lot of boilerplate in each test. I think this is a case where we'd benefit from refactoring some of the repeated code. I think we should move the creation of the mocks to a `setUp` like

```python
def setUp(self):
  self.clock_wait_event = threading.Event()
  self.mock_local_clock = mock.Mock()
  self.mock_temperature_sensor = mock.Mock()
  self.mock_temperature_store = mock.Mock()
```

These are okay to refactor because the reader can pretty much assume what they are without having to scroll up to `setUp` if they're reading the test case by itself. Something like `now` or `poll_interval` we should keep in each unit test because we're asserting based on it, so the reader wants to see what it's initialized to.

---

*[tests/test_poller.py, line 29 at r8](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KRpM9s2uUtRSGM5M1I9:-KRpM9s2uUtRSGM5M1IA:bz5zkdt) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/b17c456fad2ca1a6614a70a630b8fa277597fe78/tests/test_poller.py#L29)):*
> ```Python
> 
>         temperature_poller.start_polling_async()
>         clock_wait_event.wait(3.0)
> ```

I think it makes sense to replace the `3.0` with a named constant. It's different than the other numbers in this function in that it's not just dummy input, but a number that actually serves a purpose, so suggest something like

```python
TEST_TIMEOUT_SECONDS = 3.0
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32)*

JeetShetty:
*[tests/test_poller.py, line 15 at r8](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KRpN3oXiZPW4Vj1GESm:-KSCYV_bYL0A-sJdulfT:b-vdvbwh) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/b17c456fad2ca1a6614a70a630b8fa277597fe78/tests/test_poller.py#L15)):*
> Hmm, we're ending up with a lot of boilerplate in each test. I think this is a case where we'd benefit from refactoring some of the repeated code. I think we should move the creation of the mocks to a `setUp` like
> 
> ```python
> def setUp(self):
>   self.clock_wait_event = threading.Event()
>   self.mock_local_clock = mock.Mock()
>   self.mock_temperature_sensor = mock.Mock()
>   self.mock_temperature_store = mock.Mock()
> ```
> 
> These are okay to refactor because the reader can pretty much assume what they are without having to scroll up to `setUp` if they're reading the test case by itself. Something like `now` or `poll_interval` we should keep in each unit test because we're asserting based on it, so the reader wants to see what it's initialized to.
</details>
Did what I could with this, but it's too messy imo; ```self.mock_sensor``` and ```self.mock_store``` now become ambiguous (and ```mock_sensor``` is a misnomer in ```test_reservoir_poller```), and then ```WateringEventPoller``` is a unique case so I decided to have it not make use of ```self.mock_sensor``` or ```self.mock_store```. Maybe just ```self.clock_wait_event``` and ```self.mock_local_clock``` should go in the ```setUp```?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32)*

mtlynch:
Review status: 4 of 6 files reviewed at latest revision, 4 unresolved discussions.

---

*[tests/test_poller.py, line 15 at r8](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KRpN3oXiZPW4Vj1GESm:-KSEjqFi1Z9vBNP1cHBK:b-xaql4v) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/b17c456fad2ca1a6614a70a630b8fa277597fe78/tests/test_poller.py#L15)):*
> Did what I could with this, but it's too messy imo; ```self.mock_sensor``` and ```self.mock_store``` now become ambiguous (and ```mock_sensor``` is a misnomer in ```test_reservoir_poller```), and then ```WateringEventPoller``` is a unique case so I decided to have it not make use of ```self.mock_sensor``` or ```self.mock_store```. Maybe just ```self.clock_wait_event``` and ```self.mock_local_clock``` should go in the ```setUp```?
</details>
What do you mean when you say that `self.mock_sensor` and `self.mock_store` become ambiguous? Do you mean they represent different sensors in different tests? I think that's actually okay and embraces Python's duck typing, where objects are the "right" type as long as they do what the code needs them to do.

What about putting the tests for `WateringEventPoller` in a separate test class like

```python
class WateringEventPollerTest(unittest.TestCase):
 
    def setUp(self):
        self.clock_wait_event = threading.Event()
        self.mock_local_clock = mock.Mock()
        self.mock_pump_manager = mock.Mock()
        self.mock_watering_event_store = mock.Mock()
        self.mock_soil_moisture_store = mock.Mock()
```

If the refactoring makes it harder to follow, we can switch it back to the way that doesn't share code. For tests, it's more important that the tests be obvious and easy to read than tidy. For me, the refactored version is easier to read, but not by much, so I'll defer to what you prefer.

---

*[tests/test_poller.py, line 44 at r9](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KSEnsTDGsl-6KgorR_I:-KSEnsTDGsl-6KgorR_J:b-a02r4s) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/3c9d8468a9c0532f65d7e1a625eb6439f9b14d3e/tests/test_poller.py#L44)):*
> ```Python
>                                                 poll_interval, self.mock_sensor,
>                                                 self.mock_store)
>         now = datetime.datetime(2016, 7, 23, 10, 51, 9, 928000)
> ```

I'm trying to think of other ways we can trim down the size of these test functions without losing clarity. Two related ideas:

1. We can define the timestamp as a module-level constant so we don't have to define it in each test
1. We don't have to assign values like `humidity` to a variable

```python
TIMESTAMP_A = datetime.datetime(2016, 7, 23, 10, 51, 9, 928000)

...

def test_humidity_poller(self):
  ...
  self.mock_local_clock.now.return_value = TIMESTAMP_A
  self.mock_local_clock.wait.side_effect = (
    lambda _: self.clock_wait_event.set())
  self.mock_sensor.get_humidity_level.return_value = 50.0
 
  humidity_poller.start_polling_async()
  self.clock_wait_event.wait(TEST_TIMEOUT_SECONDS)
  self.mock_store.store_humidity.assert_called_with(TIMESTAMP_A, 50.0)
```

For (2), it's kind of counter-intuitive because we would never do that in production code because we don't want to have magic numbers, but in tests it can be helpful because we can choose numbers that are unique in the function, so it's really easy to see, "okay, 50.0 matches 50.0".

I was actually about to make this suggestion about `poll_interval` too, but I think in that case, we don't want to do it because then you'd end up with a line that looks like this:

```python
humidity_poller = poller.HumidityPoller(self.mock_local_clock,
                                                1, self.mock_sensor,
                                                self.mock_store)
```

And then it forces the reader to jump to another file to remember what the second parameter to `HumidityPoller`'s constructor is, whereas in the current implementation, it's obvious. If we really wanted to, we could do this:

```python
humidity_poller = poller.HumidityPoller(local_clock=self.mock_local_clock,
  poll_interval=1, humidity_sensor=self.mock_sensor,
  humidity_store=self.mock_store)
```

Then we solve the problem of making the `1` clear, but that feels overly verbose.

The reason we *don't* have that problem with `humidity` is that it's obvious from context what value `50.0` represents (ditto for `moisture`, `temperature`, etc.).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32)*

mtlynch:
Reviewed 2 of 2 files at r9.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32)*

JeetShetty:
*[tests/test_poller.py, line 15 at r8](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KRpN3oXiZPW4Vj1GESm:-KSNSarQfBSdn8sBY8nz:bkrd4t6) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/b17c456fad2ca1a6614a70a630b8fa277597fe78/tests/test_poller.py#L15)):*
>What do you mean when you say that self.mock_sensor and self.mock_store become ambiguous? Do you mean they represent different sensors in different tests? I think that's actually okay and embraces Python's duck typing, where objects are the "right" type as long as they do what the code needs them to do.

Yeah that is what I meant. Good to know about the duck typing!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32)*

JeetShetty:
*[tests/test_poller.py, line 44 at r9](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KSEnsTDGsl-6KgorR_I:-KSNTAydEKxrki6ZoVEI:b-786s76) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/3c9d8468a9c0532f65d7e1a625eb6439f9b14d3e/tests/test_poller.py#L44)):*
>For (2), it's kind of counter-intuitive because we would never do that in production code because we don't want to have magic numbers, but in tests it can be helpful because we can choose numbers that are unique in the function, so it's really easy to see, "okay, 50.0 matches 50.0".

Ah OK, makes sense. 

For ```poll_interval```, it seems to me like it's fine to make it a module-level constant as well. You said that we don't want to do that since we're asserting based on it, but the value isn't relevant, so someone wouldn't need to scroll up to see what the value was. But I defer to your judgment on this bc I'm still a little shaky on the reasoning of all of this. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32)*

mtlynch:
:lgtm:

---

Reviewed 1 of 1 files at r10.

---

*[tests/test_poller.py, line 44 at r9](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-KSEnsTDGsl-6KgorR_I:-KSNyZ13jR0sJ-hWIHir:b-a6je7s) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/3c9d8468a9c0532f65d7e1a625eb6439f9b14d3e/tests/test_poller.py#L44)):*
> >For (2), it's kind of counter-intuitive because we would never do that in production code because we don't want to have magic numbers, but in tests it can be helpful because we can choose numbers that are unique in the function, so it's really easy to see, "okay, 50.0 matches 50.0".
> 
> Ah OK, makes sense. 
> 
> For ```poll_interval```, it seems to me like it's fine to make it a module-level constant as well. You said that we don't want to do that since we're asserting based on it, but the value isn't relevant, so someone wouldn't need to scroll up to see what the value was. But I defer to your judgment on this bc I'm still a little shaky on the reasoning of all of this. 
</details>
Oh yeah, `poll_interval` makes sense as a module-level constant.

I might have said something confusing before about not doing it with things we're asserting on. The thing I meant to say we shouldn't do is something like this:

```python
def setUp(self):
  ...
    self.poll_interval = 1
    self.humidity_poller = poller.HumidityPoller(self.mock_local_clock,
                                                self.poll_interval, self.mock_sensor,
                                                self.mock_store)
...

def test_humidity_poller(self):
  self.mock_local_clock.now.return_value = TIMESTAMP_A
  self.mock_local_clock.wait.side_effect = (
            lambda _: self.clock_wait_event.set())
 
  self.mock_sensor.get_humidity_level.return_value = 50.0
 
  self.humidity_poller.start_polling_async()
  self.clock_wait_event.wait(TEST_TIMEOUT_SECONDS)
  self.mock_store.store_humidity.assert_called_with(TIMESTAMP_A, 50.0)
  self.mock_local_clock.wait.assert_called_with(1) # << bad
```

So basically where we hide the value that you're asserting on later because if the reader tries to read just the test function, they can't understand what's going on. It'd still be bad even if we changed the last line to:

```python
  self.mock_local_clock.wait.assert_called_with(self.poll_interval) # << bad
```

Because the reader would have to read `setUp` to understand how `self.poll_interval` is involved.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/32#-:-KSNzuu9Bf9FqQQIHUNb:bnfp4nl)*

PR #34, opened 2016-09-15
Dht11

JeetShetty:
This change is [Reviewable](https://reviewable.io/reviews/jeetshetty/greenpithumb/34)


greenpithumb/dht11.py
```diff
@@ -0,0 +1,50 @@
+import datetime
+import threading
+
+from Adafruit_DHT import common
+
+_FRESHNESS_THRESHOLD = 2
+
+
+class CachingDHT11(object):
+    """Caching DHT11 sensor wrapper.
```
mtlynch: Suggest adding as much information as can fit on the first line, so like, "Wrapper around a DHT11 that caches sensor readings."

greenpithumb/dht11.py
```diff
@@ -0,0 +1,50 @@
...
+        self._lock = threading.Lock()
+
+    def humidity(self):
+        """Returns a recent relative humidity reading."""
+        self._lock.acquire()
+        now = self._clock.now()
+        if (now - self._last_reading_time).total_seconds >= (
+                _FRESHNESS_THRESHOLD):
+            self._last_reading_time = now
+            self._last_reading = common.read(11, self._dht11_pin)
```
mtlynch: `common` as a name is kind of confusing here because it's unintuitive that refers to the DHT11 library. Suggest either

```python
from Adafruit_DHT import common as dht11
...
self._last_reading = dht11.read(11, self._dht11_pin)
```

or 

```python
import Adafruit_DHT
...
self._last_reading = Adafruit_DHT.common.read(11, self._dht11_pin)
```

greenpithumb/dht11.py
```diff
@@ -0,0 +1,50 @@
...
+
+_FRESHNESS_THRESHOLD = 2
+
+
+class CachingDHT11(object):
+    """Caching DHT11 sensor wrapper.
+
+    Reads and returns temperature and humidity levels while also caching these
+    values to ensure that the sensor is not polled at too high of a frequency.
+    """
```
mtlynch: We should mention that this class is thread-safe.

greenpithumb/dht11.py
mtlynch: We should use the defined constant (`DHT11`) so that we avoid a magic number.

greenpithumb/dht11.py
```diff
@@ -0,0 +1,50 @@
...
+        now = self._clock.now()
+        if (now - self._last_reading_time).total_seconds >= (
+                _FRESHNESS_THRESHOLD):
+            self._last_reading_time = now
+            self._last_reading = common.read(11, self._dht11_pin)
+        self._lock.release()
+        return self._last_reading[0]
+
+    def temperature(self):
+        """Returns a recent ambient temperature reading."""
```
mtlynch: `humidity` and `temperature` have a lot of similar code. I think they just differ by the last line. Can we refactor to a helper function?

mtlynch: We should say whether it's in Celsius or Farenheit.

greenpithumb/dht11.py
```diff
@@ -0,0 +1,50 @@
...
+        """
+        self._dht11_pin = dht11_pin
+        self._clock = clock
+        self._last_reading_time = datetime.min
+        self._last_reading = None
+        self._lock = threading.Lock()
+
+    def humidity(self):
+        """Returns a recent relative humidity reading."""
+        self._lock.acquire()
```
mtlynch: Ooh, good thinking on making it thread safe!

Any time we can avoid explicitly calling `acquire` and `release` it's a win because it's much harder to make a mistake that leaves `acquire` and `release` mismatched, so suggest using

```python
with self._lock:
  ...
```

Also, I think the lock is slightly too broad. It doesn't hurt anything, but just so we're thinking deliberately about it, I'm trying to work through what the lock needs to protect. The `now = self._clock.now()` can go outside the lock because it's a local variable, so each thread has its own copy and there's no risk of overwriting values in the other thread's copies. The rest *does* need the lock because they deal with instances variables, which are shared across threads so other threads could clobber our values in the middle of execution if there's no lock.

greenpithumb/dht11.py
```diff
@@ -0,0 +1,50 @@
...
+        now = self._clock.now()
+        if (now - self._last_reading_time).total_seconds >= (
+                _FRESHNESS_THRESHOLD):
+            self._last_reading_time = now
+            self._last_reading = common.read(11, self._dht11_pin)
+        self._lock.release()
+        return self._last_reading[0]
```
mtlynch: Suggest replacing the `0` here and `1` below with named constants so we avoid magic numbers.

greenpithumb/dht11.py
```diff
@@ -4,9 +4,6 @@
 from Adafruit_DHT import common
```
mtlynch: We'll need to add `Adafruit_DHT` to `requirements.txt`

greenpithumb/dht11.py
JeetShetty: >Also, I think the lock is slightly too broad. It doesn't hurt anything, but just so we're thinking deliberately about it, I'm trying to work through what the lock needs to protect. The now = self._clock.now() can go outside the lock...

It seems to me like ```now``` should stay inside the lock because ```_last_reading_time``` should be coupled to ```_last_reading```. Eg. if ```now``` is outside the lock and it takes 1 second to acquire the lock, ```_last_reading``` will appear to have been taken 1 second before it actually was.

JeetShetty: Oh I actually think this is a moo point after creating the helper function

greenpithumb/dht11.py
JeetShetty: Hmm are we adding it to requirements or adding a greenpithumb/Adafruit_DHT folder like we did with the previous dht11 code? And then add whatever packages they use to ```requirements.txt```?

mtlynch:
Reviewed 11 of 11 files at r3.
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[greenpithumb/dht11.py, line 4 at r1](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/34#-KRjgOwjXAUIP_L7edjZ-r1-4:-KSY6yMKejl-T4l7cZjI:bystujw) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/08d9649c6bf30e1d39eb2e28f7ddabd3a73e44c2/greenpithumb/dht11.py#L4)):*
> Hmm are we adding it to requirements or adding a greenpithumb/Adafruit_DHT folder like we did with the previous dht11 code? And then add whatever packages they use to ```requirements.txt```?
</details>
Bleh, I played around with it for an hour or so and it turns out it's surprisingly complicated. Normally, we'd want to just use `pip install`, but this library doesn't have an official pip package. Someone made one, but it refuses to install on a non-Raspberry Pi machine.

I think the best thing to do here is to just write `dht11.py` so that it doesn't depend on  `Adafruit_DHT` at all and push all the library-specific stuff into the `read_func` call, so callers would create a real `CachingDHT11` object like

```python
dht11 = dht11.CachingDHT11(
  lambda: Adafruit_DHT.common.read(Adafruit_DHT.common.DHT11, dht11_pin),
  clock)
```

So within `dht11.py` we can delete `self._dht11_pin` and change the line to just

```python
self._last_reading = self._dht11_read_func()
```

---

*[greenpithumb/dht11.py, line 32 at r2](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/34#-KRjgOwjXAUIP_L7edjZ-r2-32:-KSYKrlO8DkdvrIBe_b0:b9v0h7e) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/7fcb6d9b41e9abc0bce15f661863e161ccab2e0e/greenpithumb/dht11.py#L32)):*
> Oh I actually think this is a moo point after creating the helper function
</details>
Oh yeah, you're right about the `now()` call and the lock. Multi-threading is hard!

---

*[greenpithumb/dht11.py, line 39 at r2](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/34#-KRjgOwjXAUIP_L7edjZ-r2-39:-KSYL1oG2Jb2B31NzAGv:bwqu7n4) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/7fcb6d9b41e9abc0bce15f661863e161ccab2e0e/greenpithumb/dht11.py#L39)):*
> Suggest replacing the `0` here and `1` below with named constants so we avoid magic numbers.
</details>
bump on this. This doesn't depend on the `Adafruit_DHT` library because they don't define those constants either, so we should just define our own like

```python
# Position of humidity value in the tuple-returned from DHT11 read function.
_HUMIDITY_INDEX = 0
# Position of  temperature value in the tuple-returned from DHT11 read function.
_TEMPERATURE_INDEX = 1

---

*[greenpithumb/dht11.py, line 51 at r3](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/34#-KSY7z42g_h5cOfVUh66:-KSY7z42g_h5cOfVUh67:b81iczj) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/38b1f3f732aebe17b38ebbb280727045a86e0b90/greenpithumb/dht11.py#L51)):*
> ```Python
>     def humidity(self):
>         """Returns a recent relative humidity reading."""
>         with self._lock:
> ```

Suggest pushing `with self._lock` into `_read_dht11`. There's a hidden requirement in the current implementation that callers to `_read_dht11` have to be holding the lock, whereas if it's within `_read_dht11` callers don't have to care. 

---

*[greenpithumb/humidity_sensor.py, line 4 at r3](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/34#-KSYLY4MwuM9sdiS3_6T:-KSYLY4MwuM9sdiS3_6U:b-hu6ge1) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/38b1f3f732aebe17b38ebbb280727045a86e0b90/greenpithumb/humidity_sensor.py#L4)):*
> ```Python
>     """Wrapper for humidity sensor readings."""
> 
>     def __init__(self, caching_dht11):
> ```

Suggest just leaving it as `dht11`. It shouldn't really care what kind of wrappers we have around it as long as it can get the right information from it (ditto for `temperature_sensor.py`).

---

*[tests/test_dht11.py, line 10 at r3](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/34#-KSYMuWiufYr2-1jyFv2:-KSYMuWiufYr2-1jyFv3:b-12hzh) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/38b1f3f732aebe17b38ebbb280727045a86e0b90/tests/test_dht11.py#L10)):*
> ```Python
> 
> 
> class CachingDHT11Test(unittest.TestCase):
> ```

These tests are problematic in that they rely on peeking at and fiddling with `CachingDHT11`'s private members. Ideally we don't do this because we want to be able to change any of the implementation and not break our tests as long as its public interface is still doing the right thing.

What I'd suggest instead is something like

```python
    def test_gets_correct_readings_from_dht11(self):
        caching_dht11 = dht11.CachingDHT11(self.mock_dht11_read_func,
                                           self.mock_clock)
        self.mock_dht11_read_func.return_value = (50.0, 21.0)
        self.mock_clock.now.return_value = (
          datetime.datetime(2016, 1, 1, 0, 0, 0, 0))
 
        humidity = caching_dht11.humidity()
        temperature = caching_dht11.temperature()
 
        self.assertEqual(humidity, 50.0)
        self.assertEqual(temperature, 21.0)
        self.assertEqual(1, self.mock_dht11_read_func.call_count)

    def test_does_not_read_more_than_once_before_cache_expires(self):
        caching_dht11 = dht11.CachingDHT11(self.mock_dht11_read_func,
                                           self.mock_clock)
        self.mock_dht11_read_func.return_value = (50.0, 21.0)
        self.mock_clock.now.return_value = (
          datetime.datetime(2016, 1, 1, 0, 0, 0, 0))
 
        # Retrieve values for the first time.
        humidity = caching_dht11.humidity()
        temperature = caching_dht11.temperature()
 
        self.assertEqual(1, self.mock_dht11_read_func.call_count)

        humidity = caching_dht11.humidity()
        temperature = caching_dht11.temperature()

        # Set clock to just below expiration time for cached values.
        self.mock_clock.now.return_value = (
          datetime.datetime(2016, 1, 1, 0, 0, 1, 999999))

        # Call humidity() and temperature() again to verify they don't cause new
        # reads.
        caching_dht11.humidity()
        caching_dht11.temperature()
        self.assertEqual(1, self.mock_dht11_read_func.call_count)

    def test_refreshes_cache_after_cached_values_expire(self):
      # similar to last test, but when we roll the clock forward, roll it to 
      # 2 seconds later and we verify this causes the call count to go up to 2
```

* Small note: I used a different timestamp with rounder numbers just to make the changes in time more obvious.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/34)*

JeetShetty:
*[greenpithumb/dht11.py, line 39 at r2](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/34#-KRjgOwjXAUIP_L7edjZ-r2-39:-KSc4U0mHdlhHgqT-4l1:b-a65z5a) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/7fcb6d9b41e9abc0bce15f661863e161ccab2e0e/greenpithumb/dht11.py#L39)):*
> bump on this. This doesn't depend on the `Adafruit_DHT` library because they don't define those constants either, so we should just define our own like
> 
> ```python
> # Position of humidity value in the tuple-returned from DHT11 read function.
> _HUMIDITY_INDEX = 0
> # Position of  temperature value in the tuple-returned from DHT11 read function.
> _TEMPERATURE_INDEX = 1
</details>
why "tuple-returned" instead of "tuple returned"?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/34)*

JeetShetty:
*[tests/test_dht11.py, line 10 at r3](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/34#-KSYMuWiufYr2-1jyFv2:-KScGq5mVW_ThTpeNoFZ:b-bu1gns) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/38b1f3f732aebe17b38ebbb280727045a86e0b90/tests/test_dht11.py#L10)):*
> These tests are problematic in that they rely on peeking at and fiddling with `CachingDHT11`'s private members. Ideally we don't do this because we want to be able to change any of the implementation and not break our tests as long as its public interface is still doing the right thing.
> 
> What I'd suggest instead is something like
> 
> ```python
>     def test_gets_correct_readings_from_dht11(self):
>         caching_dht11 = dht11.CachingDHT11(self.mock_dht11_read_func,
>                                            self.mock_clock)
>         self.mock_dht11_read_func.return_value = (50.0, 21.0)
>         self.mock_clock.now.return_value = (
>           datetime.datetime(2016, 1, 1, 0, 0, 0, 0))
>  
>         humidity = caching_dht11.humidity()
>         temperature = caching_dht11.temperature()
>  
>         self.assertEqual(humidity, 50.0)
>         self.assertEqual(temperature, 21.0)
>         self.assertEqual(1, self.mock_dht11_read_func.call_count)
> 
>     def test_does_not_read_more_than_once_before_cache_expires(self):
>         caching_dht11 = dht11.CachingDHT11(self.mock_dht11_read_func,
>                                            self.mock_clock)
>         self.mock_dht11_read_func.return_value = (50.0, 21.0)
>         self.mock_clock.now.return_value = (
>           datetime.datetime(2016, 1, 1, 0, 0, 0, 0))
>  
>         # Retrieve values for the first time.
>         humidity = caching_dht11.humidity()
>         temperature = caching_dht11.temperature()
>  
>         self.assertEqual(1, self.mock_dht11_read_func.call_count)
> 
>         humidity = caching_dht11.humidity()
>         temperature = caching_dht11.temperature()
> 
>         # Set clock to just below expiration time for cached values.
>         self.mock_clock.now.return_value = (
>           datetime.datetime(2016, 1, 1, 0, 0, 1, 999999))
> 
>         # Call humidity() and temperature() again to verify they don't cause new
>         # reads.
>         caching_dht11.humidity()
>         caching_dht11.temperature()
>         self.assertEqual(1, self.mock_dht11_read_func.call_count)
> 
>     def test_refreshes_cache_after_cached_values_expire(self):
>       # similar to last test, but when we roll the clock forward, roll it to 
>       # 2 seconds later and we verify this causes the call count to go up to 2
> ```
> 
> * Small note: I used a different timestamp with rounder numbers just to make the changes in time more obvious.
</details>
Oh nice, I like this. Smart with the rounder numbers in the timestamp. 

I wasn't sure if your comments were for me, or if they should be in the code, but I left them in! I guess I don't really use comments which seems not good?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/34)*

mtlynch:
:lgtm:

---

Reviewed 4 of 4 files at r4.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[greenpithumb/dht11.py, line 6 at r4](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/34#-KSkNOsC0qfLU27LYR7V:-KSkNOsC0qfLU27LYR7W:bupa0j8) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/a1899f486a9545799488e464a0ac78830663bf21/greenpithumb/dht11.py#L6)):*
> ```Python
> # Maximum time a sensor reading can be used for, in seconds
> _FRESHNESS_THRESHOLD = 2
> # Position of humidity value in the tuple-returned from DHT11 read function.
> ```

Oh, sorry "tuple-returned" was a typo. Should be "tuple returned"

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/34#-:-KSkNykiV7VNlN-YgdQ_:bnfp4nl)*

PR #37, opened 2016-10-13
Update docstrings.

JeetShetty:
This change is [Reviewable](https://reviewable.io/reviews/jeetshetty/greenpithumb/37)


mtlynch:
:lgtm:

---

Reviewed 6 of 6 files at r1.

---

*[greenpithumb/pump.py, line 28 at r1](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/37#-KU02Hp8AKQDltWXoQe3:-KU02Hp8AKQDltWXoQe4:b6r3s70) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/e6ec757fe03b8a6719ed12ee14e42604fc62bed7/greenpithumb/pump.py#L28)):*
> ```Python
> 
>         Raises:
>             ValueError: The amount of water to be pumped is negative.
> ```

(very minor) I'd just say "is invalid." "negative" feels a little too specific and is the kind of thing we'd forget to update if we later decided that there's an upper limit on how much you can try to pump.

---

*[greenpithumb/pump.py, line 30 at r1](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/37#-KU02Cap8-qi1_25aupW:-KU02Cap8-qi1_25aupX:br8jfjk) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/e6ec757fe03b8a6719ed12ee14e42604fc62bed7/greenpithumb/pump.py#L30)):*
> ```Python
>             ValueError: The amount of water to be pumped is negative.
>         """
> ```

(out of scope, but) Extraneous newline.

---

*[greenpithumb/pump_manager.py, line 36 at r1](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/37#-KU031BHI32Do0YJxyVU:-KU031BHI32Do0YJxyVV:b-q24f1i) ([raw file](https://github.com/jeetshetty/greenpithumb/blob/e6ec757fe03b8a6719ed12ee14e42604fc62bed7/greenpithumb/pump_manager.py#L36)):*
> ```Python
> 
>         Returns:
>             An int of the amount of water pumped, in mL.
> ```

I don't know what the accepted wisdom on this is, but I try to minimize references to actual types in Python docstrings because it feels at odds with the language's flexible typing.

I think this is a case where we'd want to avoid saying `int` because that's not a really important part of the interface and we might decide for whatever reason to change the default pump amount to a fractional value and callers shouldn't care.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jeetshetty/greenpithumb/37#-:-KU03rzWkDWU9erLe6aR:bnfp4nl)*

PR #38, opened 2016-12-01
Implement IO class.

JeetShetty:
This change is [Reviewable](https://reviewable.io/reviews/jeetshetty/greenpithumb/38)


greenpithumb/io.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def __init__(self, placeholder_gpio):
+        """Creates a new input/output wrapper.
+
+        Args:
+            placeholder_gpio: Placeholder for RPi.GPIO.
+        """
+        self.GPIO = placeholder_gpio
+
+    def write_pin(self, channel, value):
```
mtlynch: I think we should say `pin` instead of `channel`. The ADC's inputs are channels, but the GPIO pins are just pins.

greenpithumb/light_sensor.py
```diff
@@ -15,14 +13,16 @@ class LightSensorLowError(Error):
...
     """Wrapper for light sensor."""
 
-    def __init__(self, adc):
+    def __init__(self, adc, pin):
         """Creates a new LightSensor wrapper.
 
         Args:
             adc: ADC(analog to digital) interface to receive
                 analog signals from light sensor.
+            pin: ADC pin to which the light sensor is connected.
```
mtlynch: So this one should be channel

greenpithumb/weight_sensor.py
```diff
@@ -1,21 +1,20 @@
-import adc
-
-
 class WeightSensor(object):
     """Wrapper for a weight sensor."""
 
-    def __init__(self, adc):
+    def __init__(self, adc, pin):
```
mtlynch: ditto about pin -> channel

greenpithumb/io.py
```diff
@@ -0,0 +1,55 @@
+#import RPi.GPIO as GPIO
```
mtlynch: We might want to avoid the `io` name because Python already has a built in library with that name:

https://docs.python.org/2/library/io.html

pi_io?

greenpithumb/io.py
```diff
@@ -0,0 +1,55 @@
...
+            channel: Index of Raspberry Pi channel to write to.
+            value: Value to write to the pin (must be either 0 or 1023).
+
+        Raises:
+            WriteError: The write value is invalid.
+        """
+        if not (value == SIGNAL_OFF or value == SIGNAL_ON):
+            raise WriteError(
+                'Write value of %s is invalid. Must be either %s or %s' %
+                (value, SIGNAL_OFF, SIGNAL_ON))
```
JeetShetty: smart!

greenpithumb/pi_io.py
```diff
@@ -0,0 +1,43 @@
...
+        self._GPIO = gpio
+
+    def turn_pin_on(self, pin):
+        """Turns on a Raspberry Pi GPIO pin.
+
+        Args:
+            pin: Index of Raspberry Pi pin to turn on
+        """
+        self._GPIO.setup(pin, self._GPIO.OUT)
+        self._GPIO.output(pin, 1)
```
mtlynch: I think we want `self._GPIO.HIGH` instead of `1` and `self._GPIO.LOW` instead of `0`

greenpithumb/pi_io.py
```diff
@@ -0,0 +1,43 @@
+#import RPi.GPIO as GPIO
+
+# TODO(mtlynch): Move these pin constants to a configuration file so they're not
+# hardcoded in.
+
+# Raspberry Pi pin to which each component is connected
+PIN_PUMP = 0
```
mtlynch: (kind of out of scope, but since we're touching it): I think we should take this out of here and have the caller to `Pump` specify the pump's pin in the constructor. It would match all the ADC components where the caller has to specify the ADC channel in the constructor.

greenpithumb/pi_io.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def turn_pin_on(self, pin):
+        """Turns on a Raspberry Pi GPIO pin.
+
+        Args:
+            pin: Index of Raspberry Pi pin to turn on
+        """
+        self._GPIO.setup(pin, self._GPIO.OUT)
+        self._GPIO.output(pin, 1)
+        self._GPIO.cleanup(pin)
```
mtlynch: Calling `GPIO.cleanup` turns off all the GPIO pins and I think prevents us from calling `GPIO` for the rest of the program, so we only want to do that at the end.

Probably the best way to address is to have a `close` method that calls `cleanup`. We can clean it up later to use context managers (or something like [contextlib.closing](https://docs.python.org/2.7/library/contextlib.html)) but just having a `close` method and documenting that the caller has to call it when done should be enough for now.

Also, the caller should only ever create one of these instances, but we'll just document that and look at ways to clean it up later.

greenpithumb/pi_io.py
```diff
@@ -0,0 +1,43 @@
...
+        """
+        self._GPIO = gpio
+
+    def turn_pin_on(self, pin):
+        """Turns on a Raspberry Pi GPIO pin.
+
+        Args:
+            pin: Index of Raspberry Pi pin to turn on
+        """
+        self._GPIO.setup(pin, self._GPIO.OUT)
```
mtlynch: Once we call `setup` to set a pin as output, we don't have to call it again. I don't think it hurts anything, but it's a bit messy. We could maintain a `set` of all the pins we've already set up like

```python
def __init__(self, gpio):
  self._GPIO = gpio
  self._output_pins = set()
...

def ensure_pin_is_output(self, pin):
  if pin in self._output_pins:
    return
  self._GPIO.setup(pin, self._GPIO.OUT)
  self._output_pins.add(pin)

def turn_pin_on(self, pin):
  ensure_pin_is_output(pin)
  ...
```

greenpithumb/pi_io.py
```diff
@@ -21,23 +16,36 @@ def __init__(self, gpio):
...
+        """Adds pin to output pin set if it is not already in it."""
+        if pin in self._output_pins:
+            return
         self._GPIO.setup(pin, self._GPIO.OUT)
-        self._GPIO.output(pin, 0)
-        self._GPIO.cleanup(pin)
\ No newline at end of file
+        self._output_pins.add(pin)
+
+    def _close(self):
```
mtlynch: `close` needs to be public because the caller needs access to it

tests/test_pi_io.py
```diff
@@ -14,14 +14,30 @@ def test_turn_pin_on(self):
...
+        IO.turn_pin_off(pin_one)
+        IO.turn_pin_on(pin_one)
+        IO.turn_pin_off(pin_two)
+        IO.turn_pin_on(pin_two)
+
+        self.assertEqual(mock_GPIO.setup.call_args_list,
+                         [((pin_one, mock_GPIO.OUT),),
+                          ((pin_two, mock_GPIO.OUT),)])
+        pin_list = [pin_one, pin_two]
+        self.assertEqual(IO._output_pins, set(pin_list))
```
mtlynch: Suggest cutting 42-43.

39-41 gets us sufficient confidence that the class did the right thing. We're interested in whether `IO` correctly calls the `GPIO` library. The `_output_pins` is an implementation detail that the test should ideally not peek into.

tests/test_pump.py
```diff
@@ -13,7 +12,8 @@ class PumpTest(unittest.TestCase):
     def setUp(self):
         self.mock_pi_io = mock.Mock()
         self.mock_clock = mock.Mock(spec=clock.Clock)
-        self.pump = pump.Pump(self.mock_pi_io, self.mock_clock)
+        self.pump_pin = 0
```
mtlynch: Suggest using a different value to make it clearer that this is a GPIO pin. On a Pi3, these can be 2 to 27. Plus I'm generally nervous about using `0` because I feel like there's potential for the test to pass by mistake, since it's easier for something to accidentally get set to `0` than say `6`.

tests/test_pump.py
```diff
@@ -26,8 +26,8 @@ def test_pump_4_3_liters(self):
         # verify that the values are equal to several significant digits.
         self.assertAlmostEqual(wait_time_expected, wait_time_actual)
         # Verify the signal gets sent to turn the pump pin on, then off.
-        self.mock_pi_io.turn_pin_on.assert_called_once_with(pi_io.PIN_PUMP)
-        self.mock_pi_io.turn_pin_off.assert_called_once_with(pi_io.PIN_PUMP)
+        self.mock_pi_io.turn_pin_on.assert_called_once_with(self.pump_pin)
```
mtlynch: Ideally the reader shouldn't have to read the `setUp` function to understand why a test passes. For things like `self.mock_clock` I feel like it's easy for the reader to assume it's the pump's internal clock. For things like the GPIO pin number where it's set to a particular value, I feel like it's less easy to assume and the reader has to go to the `setUp` to understand what's going on.

Suggest instantiating the pump in each test function like:

```python
pump.Pump(self.mock_pi_io, self.mock_clock, pump_pin=5).pump_water(4300.0)
...
self.mock_pi_io.turn_pin_on.assert_called_once_with(5)
self.mock_pi_io.turn_pin_off.assert_called_once_with(5)
```

That way, the reason for the pin being a particular value is obvious within the test function itself.

tests/test_pump.py
```diff
@@ -26,8 +26,8 @@ def test_pump_4_3_liters(self):
...
+        self.mock_pi_io.turn_pin_off.assert_called_once_with(self.pump_pin)
```
mtlynch: (optional) Consider using [`assert_has_calls`](https://docs.python.org/dev/library/unittest.mock.html#unittest.mock.Mock.assert_has_calls) because it enforces an ordering to the calls. But if it's a pain to get the syntax right, I think the current assertions are fine.

JeetShetty: How do we enforce ordering for two different methods?

tests/test_pi_io.py
```diff
@@ -39,5 +39,3 @@ def test_setup_only_called_once_per_pin(self):
         self.assertEqual(mock_GPIO.setup.call_args_list,
                          [((pin_one, mock_GPIO.OUT),),
                           ((pin_two, mock_GPIO.OUT),)])
```
mtlynch: Oh we should add a test case to make sure `close` works. I think really simple like

```python
     def test_close(self):
         mock_GPIO = mock.Mock()
         io = pi_io.IO(mock_GPIO)
         io.close()
         mock_GPIO.cleanup.assert_called_once()
```

tests/test_pi_io.py
```diff
@@ -39,5 +39,3 @@ def test_setup_only_called_once_per_pin(self):
         self.assertEqual(mock_GPIO.setup.call_args_list,
```
mtlynch: Oh, I didn't notice before but `IO` should be `io` otherwise it looks like a class:
https://google.github.io/styleguide/pyguide.html?showone=Naming#Naming

PR #39, opened 2016-12-01
Add DbRetriever class.

JeetShetty:
This change is [Reviewable](https://reviewable.io/reviews/jeetshetty/greenpithumb/39)


greenpithumb/db_retriever.py
```diff
@@ -0,0 +1,40 @@
...
+            temperature information.
+        """
+        self._cursor.execute('''SELECT * FROM temperature WHERE timestamp >= ?
+                             and timestamp <= ?''' (range_start, range_end))
+        temperature_data = self._cursor.fetchall()
+
+        temperature_data_list = []
+        for row in temperature_data:
+            d = {}
+            d['timestamp'] = row[0]
```
mtlynch: I think we should parse these back into Python objects. The frontend will eventually convert it to JSON and the code here is convenient because it's very close to JSON, but I think we should keep it as an object for as long as possible so that this library stays generic and the frontend is what handles JSON stuff.

So, like:

```python
class TemperatureRecord(object):

  def __init__(self, temperature, timestamp):
    self._temperature = temperature
    self._timestamp = timestamp

  @property
  def temperature(self):
    return self._temperature

  @property
  def timestamp(self):
    return self._timestamp
```

But that's going to get super tedious for all the different record types, so I think we can use named tuples:

```python
TemperatureRecord = collections.namedtuple('TemperatureRecord', ['temperature', 'timestamp'])
```

https://docs.python.org/2/library/collections.html#namedtuple-factory-function-for-tuples-with-named-fields
(the documentation also specifically talks about deserializing from sqlite)

Also, we should probably update the code in the `store_` functions to explicitly convert dates to strings. I suspect the sqlite3 library probably just does the conversion for us silently, but we should do it explicitly because otherwise our deserialization logic depends on sqlite's serialization logic that we don't really see (and theoretically could change).

JeetShetty: Oooh namedtuples fancy. 

Agreed on explicitly converting dates to strings. I'll do that in a separate PR.

tests/test_db_store.py
```diff
@@ -34,6 +34,20 @@ def test_latest_soil_moisture_empty_database(self):
...
+        timestamp1 = str(datetime.datetime(2016, 7, 23, 10, 51, 9, 928000))
+        timestamp2 = str(datetime.datetime(2016, 7, 23, 10, 52, 9, 928000))
+        mock_cursor.fetchall.return_value = [(timestamp1, 300),
+                                             (timestamp2, 400)]
+        soil_moisture_data = store.retrieve_soil_moisture()
+
+        self.assertEqual(soil_moisture_data[0].timestamp, timestamp1)
+        self.assertEqual(soil_moisture_data[0].soil_moisture, 300)
+        self.assertEqual(soil_moisture_data[1].timestamp, timestamp2)
+        self.assertEqual(soil_moisture_data[1].soil_moisture, 400)
```
mtlynch: One subtlety here is that the test assumes an ordering or the result list, but the function contract doesn't guarantee one. The actual implementation is that the order is just the same as the order from sqlite, but that's an implementation detail and the test shouldn't depend on that.

Normally if we wanted to compare lists without caring about order, we could just use `assertItemsEqual`, but it's a pain because I don't think our `namedtuple`s implement `equals` the way we want.

I think that leaves either enforcing ordering as part of the function contract (e.g. "returns soil moisture and timestamp in chronological order") or leave the function as is and just sort the results in the test so we're guaranteed an ordering.

tests/test_db_store.py
```diff
@@ -34,6 +34,20 @@ def test_latest_soil_moisture_empty_database(self):
...
+    def test_retrieve_soil_moisture(self):
+        mock_cursor = mock.Mock()
+        store = db_store.SoilMoistureStore(mock_cursor)
+        timestamp1 = str(datetime.datetime(2016, 7, 23, 10, 51, 9, 928000))
+        timestamp2 = str(datetime.datetime(2016, 7, 23, 10, 52, 9, 928000))
+        mock_cursor.fetchall.return_value = [(timestamp1, 300),
+                                             (timestamp2, 400)]
+        soil_moisture_data = store.retrieve_soil_moisture()
+
+        self.assertEqual(soil_moisture_data[0].timestamp, timestamp1)
```
JeetShetty: Mm this and the above comment are interesting to me. The reason that I tend to prefer using ```timestamp``` variables is that for me my eyes tend to glaze over when I'm looking at ```datetime``` objects (or any "long" objects) and I think I'd find it difficult reading someone else's tests to figure out what's supposed to be matching up with what, whereas with named variables it's clear. I think I'd prefer using naming like ```string_timestamp1``` or something to clear up the ambiguity. I suppose another good option is to use simpler ```datetime``` values that are more readable, which I think might have already been your suggestion on a previous PR.

tests/test_db_store.py
```diff
@@ -34,6 +34,20 @@ def test_latest_soil_moisture_empty_database(self):
         moisture = store.latest_soil_moisture()
         self.assertIsNone(moisture)
 
+    def test_retrieve_soil_moisture(self):
+        mock_cursor = mock.Mock()
+        store = db_store.SoilMoistureStore(mock_cursor)
+        timestamp1 = str(datetime.datetime(2016, 7, 23, 10, 51, 9, 928000))
```
JeetShetty: Hmm so I'm highly confused about ```datetime``` and converting to strings and back, bc I guess ```strftime``` and ```strptime``` don't like time zones?! What's the Z above supposed to be? And how does that convert to pytz.utc when you make it into a ```datetime.datetime``` object?

greenpithumb/db_store.py
```diff
@@ -41,17 +41,20 @@ def latest_soil_moisture(self):
...
+        self._cursor.execute('SELECT * FROM soil_moisture')
         SoilMoistureRecord = collections.namedtuple(
             'SoilMoistureRecord', ['timestamp', 'soil_moisture'])
         soil_moisture_data = map(SoilMoistureRecord._make,
                                  self._cursor.fetchall())
-        return soil_moisture_data
+        if soil_moisture_data == []:
+            return None
+        else:
+            return soil_moisture_data
```
mtlynch: (optional) This is just a style thing but I like avoiding nesting when I can, so if I ever have a thing of like 

```python
if whatever:
  return x
else:
  return y
```

I'd rather just do

```python:
if whatever:
  return x
return y
```

greenpithumb/db_store.py
```diff
@@ -41,17 +41,20 @@ def latest_soil_moisture(self):
...
+            A list of objects with 'timestamp' and 'soil_moisture' fields.
         """
-        self._cursor.execute("SELECT * FROM soil_moisture")
+        self._cursor.execute('SELECT * FROM soil_moisture')
         SoilMoistureRecord = collections.namedtuple(
             'SoilMoistureRecord', ['timestamp', 'soil_moisture'])
         soil_moisture_data = map(SoilMoistureRecord._make,
                                  self._cursor.fetchall())
-        return soil_moisture_data
+        if soil_moisture_data == []:
```
mtlynch: We can use implicit false here so just

```python
if not soil_moisture_data:
```

https://google.github.io/styleguide/pyguide.html?showone=True/False_evaluations#True/False_evaluations

greenpithumb/db_store.py
```diff
@@ -41,17 +41,20 @@ def latest_soil_moisture(self):
...
+            return None
```
mtlynch: Suggest returning `[]` here since the library is easier to use if we guarantee the result is always a list (like caller can always iterate it, always check the length).

greenpithumb/db_store.py
```diff
@@ -51,10 +51,7 @@ def retrieve_soil_moisture(self):
             'SoilMoistureRecord', ['timestamp', 'soil_moisture'])
         soil_moisture_data = map(SoilMoistureRecord._make,
                                  self._cursor.fetchall())
-        if soil_moisture_data == []:
-            return None
-        else:
-            return soil_moisture_data
+        return soil_moisture_data
```
mtlynch: This is the only outstanding note:

>I think we should make the timestamp field be a datetime because store_soil_moisture takes timestamps in the form of datetimes so it's confusing for retrieve_soil_moisture to give it back in the form of a string.

What do you want to do about that one?

tests/test_db_store.py
```diff
@@ -3,6 +3,7 @@
 
 import mock
 import datetime
```
mtlynch: `datetime` should be in the first group of imports because it's a standard library module
https://google.github.io/styleguide/pyguide.html?showone=Imports_formatting#Imports_formatting

tests/test_db_store.py
```diff
@@ -38,18 +39,32 @@ def test_retrieve_soil_moisture(self):
...
-                         '2016-07-23 10:51:09.928000')
+        self.assertEqual(
+            soil_moisture_data[0].timestamp,
+            datetime.datetime(
+                2016,
+                7, 23,
+                10,
+                51,
+                9, 928000,
+                tzinfo=tzoffset(None, -18000)))
```
mtlynch: I wish we had a better solution for this. I feel like it's hard for the reader to recognize that `-05:00` matches up to `tzoffset(None, -18000)`.

I have two ideas, but I don't love either of them.

### 1. Use UTC

We can change the `-05:00` part to `+00:00` (UTC) or `Z` (shorthand for UTC), then in the assertion, change it to `tzinfo.pytz.utc`. It's a little easier to identify, but it obscures the fact that the database stores times in local time instead of UTC.

### 2. Use helper function for time zone

We can alternatively replace the `tzoffset` call with a global constant like:

```python
# Timezone offset info for EST (UTC minus 5 hours).
UTC_MINUS_5 = tzoffset(None, -18000)
...

        self.assertEqual(
            soil_moisture_data[1].timestamp,
            datetime.datetime(
                2016,
                7, 23,
                10,
                52,
                9, 928000,
                tzinfo=UTC_MINUS_5))
```

What do you think?

tests/test_db_store.py
```diff
@@ -3,6 +3,7 @@
 
 import mock
 import datetime
+from dateutil.tz import tzoffset
```
mtlynch: Style guide forbids importing functions directly, so we should either do `from dateutil import tz` or `import dateutil.tz`
https://google.github.io/styleguide/pyguide.html?showone=Imports#Imports

tests/test_db_store.py
JeetShetty: Oh I like the global constant! Why don't you love that option?

PR #45, opened 2017-01-03
Convert timestamps to strings before storing in DB.

JeetShetty:
This change is [Reviewable](https://reviewable.io/reviews/jeetshetty/greenpithumb/45)


greenpithumb/db_store.py
```diff
@@ -26,7 +26,7 @@ def store_soil_moisture(self, timestamp, soil_moisture):
             soil_moisture: An int of the soil moisture reading.
         """
         self._cursor.execute('INSERT INTO soil_moisture VALUES (?, ?)',
-                             (timestamp, soil_moisture))
+                             (str(timestamp), soil_moisture))
```
mtlynch: Can we create a function like `_serialize_timestamp` that takes a `datetime` and returns a string? That way all the functions share a single implementation and we can change the format easily in one spot.

I think we should do `timestamp.isoformat` because it's more obvious how it's serializing it. I wouldn't know off the top of my head that `str(timestamp)` serializes it in ISO format unless I jumped to the documentation or tests.

greenpithumb/db_store.py
```diff
@@ -2,6 +2,18 @@
 from dateutil import parser
 
 
+def _serialize_timestamp(timestamp):
+    """Converts a timestamp to a string.
+
+    Args:
+        timestamp: Timestamp in datetime.datetime format.
```
mtlynch: Suggest just saying "Timestamp as a datetime object." The "format" makes it sound like it's already a string.

greenpithumb/db_store.py
```diff
@@ -2,6 +2,18 @@
...
+
+    Returns:
+        Timestamp as a string in ISO 8601 format.
+    """
+    return timestamp.isoformat(' ')
```
mtlynch: Suggest using `'T'` a the separator character because that's most common for ISO 8601 strings.

PR #46, opened 2017-01-05
Add image capture from a Raspberry Pi camera.

JeetShetty:
This change is [Reviewable](https://reviewable.io/reviews/jeetshetty/greenpithumb/46)


greenpithumb/camera.py
```diff
@@ -0,0 +1,27 @@
...
+            local_clock: Local clock interface.
+            camera: Camera interface.
+        """
+        self._folder_path = folder_path
+        self._local_clock = local_clock
+        self._camera = camera
+
+    def capture(self):
+        """Captures an image from the camera."""
+        filename = self._local_clock.now().strftime('%Y-%m-%d-%H')
```
mtlynch: I think we should do the full timestamp to seconds-level granularity so that we can support taking many pictures per hour or minute. Also, since we're using the filename to store the metadata about the time the picture was taken, I think we should include the UTC offset so that we can parse it back later as a timezone-aware `datetime`.

One thing to be aware of is that `:` is an illegal character on some filesystems, so we don't want to use that in our filename.

Also I don't think `picamera` will add the extension for us, so I think we need to append `'.jpg'`.

greenpithumb/camera.py
```diff
@@ -0,0 +1,27 @@
+class Camera(object):
```
mtlynch: I'd say it's more like a `CameraManager` or a `PhotoSaver` because the only thing the caller can do with it is tell it to save a picture to the filesystem.

greenpithumb/camera.py
```diff
@@ -0,0 +1,27 @@
...
+                will be stored.
+            local_clock: Local clock interface.
+            camera: Camera interface.
+        """
+        self._folder_path = folder_path
+        self._local_clock = local_clock
+        self._camera = camera
+
+    def capture(self):
+        """Captures an image from the camera."""
```
mtlynch: Suggests "Captures an image from the camera and saves it to the destination folder." Maybe rename the function to something like `save_photo` to make the purpose a little clearer.

greenpithumb/poller.py
```diff
@@ -190,3 +190,23 @@ def _poll_once(self):
             if ml_pumped > 0:
                 self._watering_event_store.store_water_pumped(
                     self._local_clock.now(), ml_pumped)
+
+
+class CameraPoller(SensorPollerBase):
```
mtlynch: Oh this is clever! It hadn't occurred to me that it would just plug into our pollers.

greenpithumb/camera.py
```diff
@@ -0,0 +1,27 @@
...
+        filename = self._local_clock.now().strftime('%Y-%m-%d-%H')
+        self._camera.capture(self._folder_path + filename, format='jpeg')
```
mtlynch: We always want to use [`os.path.join`](https://docs.python.org/2/library/os.path.html#os.path.join) to combine file paths. There's a lot of subtle things that can go wrong if we do it by hand (especially if we want to work well across platforms). A common bug is to call the function and forget the trailing slash like `C:\foo\bar` instead of `C:\foo\bar\` and then it puts the file at `C:\foo\bar2016-01-05.jpeg`.

tests/test_camera.py
```diff
@@ -0,0 +1,25 @@
+import unittest
+import datetime
```
mtlynch: `datetime` should be first

tests/test_camera.py
```diff
@@ -0,0 +1,25 @@
...
+    def test_capture(self):
+        folder_path = 'C:\\Users\\Jeet\\'
+        mock_local_clock = mock.Mock()
+        mock_camera = mock.Mock()
+        mock_local_clock.now.return_value = datetime.datetime(2016, 7, 23, 10,
+                                                              51, 9, 928000)
+        camera.Camera(folder_path, mock_local_clock, mock_camera).capture()
+        mock_camera.capture.assert_called_once_with(
+            'C:\\Users\\Jeet\\2016-07-23-10',
+            format='jpeg')
```
mtlynch: Suggest using [`contextlib.closing`](https://docs.python.org/2.7/library/contextlib.html#contextlib.closing) here. It calls `close` automatically at the end of the `with` block, even if there's an exception.

That's what we'll want actual callers to use.

```python
        with contextlib.closing(camera.Camera(folder_path, mock_local_clock,
                                              mock_camera)) as camera_manager:
            camera_manager.capture()

        mock_camera.capture.assert_called_once_with(
            'C:\\Users\\Jeet\\2016-07-23-10',
            format='jpeg')
        mock_camera.close.assert_called()
```

tests/test_camera_manager.py
```diff
@@ -0,0 +1,31 @@
...
+        mock_local_clock = mock.Mock()
+        mock_camera = mock.Mock()
+        mock_local_clock.now.return_value = datetime.datetime(
+            2016, 7, 23, 10, 51, 9,
+            928000, tzinfo=pytz.utc)
+        with contextlib.closing(camera_manager.CameraManager(
+                folder_path, mock_local_clock, mock_camera)) as manager:
+            manager.save_photo()
+            mock_camera.capture.assert_called_once_with(
+                'C:/Users/Jeet/2016-07-23T10-51-09.928000+00-00.jpg')
```
mtlynch: I think this should be just one test, like we do the `mock_camera.close.assert_called()` outside the with. I feel like in this case, it's still all logically part of the same single scenario of taking a photo and releasing the camera manager.

Issue #43, opened 2016-12-31
Integrate GPIO stuff in moisture sensor


mtlynch:
I realized the code for `moisture_sensor.py` omits some critical elements we need. The moisture sensor is powered by the GPIO pins, because I think if we leave it powered all the time, the copper starts to corrode.

We can see how Dickson Chow does it here:

https://github.com/dicksondickson/PlantFriends/blob/7c172e9f80c2de2fdfc373d31a89de66787a7756/sensor_node/RFM69/sensor_node.ino#L306...L329

So it looks like what we need to do is:

1. Set one pin to high, one pin to low
1. Read the sensor value from the ADC
1. Set both pins to low
1. Wait for capacitive effects to go away (he waits 88 milliseconds)
1. Repeat 1-2, but with the pins swapped and inverting the value from 1023
1. Turn off both pins
1. Average the two readings as the moisture level

JeetShetty:
What pins are we writing to in step 1? I'm confused about how this works through the ADC.

mtlynch:
So there 6 GPIO pins in play here. The ADC needs 4 pins, but the `moisture_sensor` module doesn't care about those 4 pins because it just needs an ADC instance (the instance worries about those pins and abstracts it away from `moisture_sensor`).

BUT `moisture_sensor` needs to control power to the sensor, so it needs to use two GPIO pins as output so that it can turn power on and off to the moisture sensor. Then it needs to know which ADC channel the moisture sensor is plugged into. In our schematic, the moisture sensor is plugged into ADC channel 7.

This is the quick 'n dirty script I used to test the soil moisture sensor as a standalone program. The numbers match the Fritzing schematic we have for the wiring.

```python
import time

import Adafruit_MCP3008
import RPi.GPIO as GPIO

# mtlynch: The moisture_sensor module wouldn't need to care about this stuff because it'd
# have an ADC initialized already
CLK  = 18
MISO = 23
MOSI = 24
CS   = 25
mcp = Adafruit_MCP3008.MCP3008(clk=CLK, cs=CS, miso=MISO, mosi=MOSI)

# Below this line is going to be kind of similar to what we want to do in moisture_sensor
ADC_CHANNEL_MOISTURE_SENSOR = 7

GPIO_PIN_1 = 16
GPIO_PIN_2 = 12

GPIO.setmode(GPIO.BCM)
GPIO.setup(GPIO_PIN_1, GPIO.OUT)
[...]

PR #53, opened 2017-01-10
Update MoistureSensor to use GPIO pins.

JeetShetty:
* Issue [#43](https://github.com/JeetShetty/GreenPiThumb/issues/43)

---
This change is [Reviewable](https://reviewable.io/reviews/jeetshetty/greenpithumb/53)


greenpithumb/moisture_sensor.py
```diff
@@ -1,17 +1,44 @@
...
-    def __init__(self, adc, channel):
+    def __init__(self, adc, pi_io, channel, gpio_pins, clock):
         """Creates a new MoistureSensor wrapper.
 
         Args:
             adc: ADC(analog to digital) interface to receive analog signals from
                 moisture sensor.
+            pi_io: Raspberry Pi I/O interface.
             channel: ADC channel to which the moisture sensor is connected.
+            gpio_pins: Tuple of length 2 of the Raspberry Pi GPIO pins that
```
mtlynch: I feel like we might as well just make them two ints. Tuple would be more useful if it might eventually grow to > 2 pins, but this is a case where we know for sure it will always be 2. We save ourselves a variable here, but I think we add complexity because with a tuple, it can have a length != 2 and we crash, whereas the caller can't screw up two ints.

greenpithumb/moisture_sensor.py
```diff
@@ -1,17 +1,44 @@
...
+        self._pi_io.turn_pin_off(pin_two)
+        moisture_one = self._adc.read_adc(self._channel)
+
+        self._clock.wait(0.1)
+
+        self._pi_io.turn_pin_on(pin_two)
+        self._pi_io.turn_pin_off(pin_one)
+        moisture_two = 1023 - self._adc.read_adc(self._channel)
+
+        self._pi_io.turn_pin_off(pin_two)
```
mtlynch: It might be a good idea to guarantee we always leave both pins in an off state, even if `read_adc` throws an exception (we can use the `finally` keyword for this). Can we add a unit test for this, too?

tests/test_moisture_sensor.py
```diff
@@ -7,14 +7,21 @@
 
 class MoistureSensorTest(unittest.TestCase):
 
-    def setUp(self):
-        self.mock_adc = mock.Mock()
+    def test_moisture_50(self):
```
mtlynch: The `50`'s out of date, but I think we can do something more descriptive, like `test_moisture_read_succeeds_when_adc_succeeds`.

tests/test_moisture_sensor.py
```diff
@@ -7,14 +7,21 @@
...
-    def test_moisture_50(self):
-        """Moisture level read should be moisture level returned."""
-        self.mock_adc.read_adc.return_value = 50
-        moisture_level = self.moisture_sensor.moisture()
-        self.assertEqual(moisture_level, 50)
+        self.assertIn(mock.call(12), mock_pi_io.turn_pin_on.call_args_list)
+        self.assertIn(mock.call(12), mock_pi_io.turn_pin_off.call_args_list)
+        self.assertIn(mock.call(16), mock_pi_io.turn_pin_on.call_args_list)
+        self.assertIn(mock.call(16), mock_pi_io.turn_pin_off.call_args_list)
+        mock_clock.wait.assert_called_once()
```
mtlynch: Might as well assert it got the time right.

tests/test_moisture_sensor.py
```diff
@@ -7,14 +7,21 @@
...
+        mock_adc.read_adc.side_effect = [300, (1023 - 400)]
+        mock_pi_io = mock.Mock()
+        mock_clock = mock.Mock()
         channel = 1
-        self.moisture_sensor = moisture_sensor.MoistureSensor(self.mock_adc,
-                                                              channel)
+        gpio_pins = (12, 16)
+        sensor = moisture_sensor.MoistureSensor(mock_adc, mock_pi_io, channel,
+                                                gpio_pins, mock_clock)
+        moisture_level = sensor.moisture()
```
mtlynch: Suggest reducing this to just `self.assertEqual(50, sensor.moisture())`

tests/test_moisture_sensor.py
```diff
@@ -7,14 +7,21 @@
...
 
-    def test_moisture_50(self):
-        """Moisture level read should be moisture level returned."""
-        self.mock_adc.read_adc.return_value = 50
-        moisture_level = self.moisture_sensor.moisture()
-        self.assertEqual(moisture_level, 50)
+        self.assertIn(mock.call(12), mock_pi_io.turn_pin_on.call_args_list)
+        self.assertIn(mock.call(12), mock_pi_io.turn_pin_off.call_args_list)
+        self.assertIn(mock.call(16), mock_pi_io.turn_pin_on.call_args_list)
+        self.assertIn(mock.call(16), mock_pi_io.turn_pin_off.call_args_list)
```
mtlynch: The problem here is that 22 and 23 are redundant, as are 24 and 25 (it would still pass with only one call). And it hides the fact that we're actually calling `turn_pin_off` three times.

Suggest checking like this:

```python
        self.assertEqual([mock.call(12), mock.call(16)],
                         mock_pi_io.turn_pin_on.call_args_list)
        self.assertEqual([mock.call(16), mock.call(12), mock.call(16)],
                         mock_pi_io.turn_pin_off.call_args_list)
```

tests/test_moisture_sensor.py
```diff
@@ -7,21 +7,38 @@
...
+        channel = 1
+        gpio_pin_1 = 12
+        gpio_pin_2 = 16
+        sensor = moisture_sensor.MoistureSensor(
+            mock_adc, mock_pi_io, channel, gpio_pin_1, gpio_pin_2, mock_clock)
+
+        with self.assertRaises(ValueError):
+            sensor.moisture()
+            self.assertEqual([mock.call(16), mock.call(12), mock.call(16)],
+                             mock_pi_io.turn_pin_off.call_args_list)
```
mtlynch: (optional) I'd like it to be more explicit that we're testing that the pins be left in an "off" state. The way we're currently testing is kind of roundabout because we're verifying that a particular sequence was called, but it's possible for the implementation to call that sequence and still leave the pins in an "on" state.

Here's an idea, but I'm not 100% happy with it because it really should be something we have in both tests (because it should leave pins in off state no matter what) but it's kind of a large amount of code to copy/paste both places, but I don't have any good ideas about how to share the code between both tests.

```python
    def test_moisture_sensor_pins_turned_off_if_adc_raises_exception(self):
        mock_adc = mock.Mock()
        mock_adc.read_adc.side_effect = ValueError()
        pin_state = collections.defaultdict(lambda: False)

        def mock_turn_pin_on(pin):
            pin_state[pin] = True

        def mock_turn_pin_off(pin):
            pin_state[pin] = False

        mock_pi_io = mock.Mock()
        mock_pi_io.turn_pin_on.side_effect = mock_turn_pin_on
        mock_pi_io.turn_pin_off.side_effect = mock_turn_pin_off
        mock_clock = mock.Mock()
        channel = 1
        gpio_pin_1 = 12
        gpio_pin_2 = 16
        sensor = moisture_sensor.MoistureSensor(
            mock_adc, mock_pi_io, channel, gpio_pin_1, gpio_pin_2, mock_clock)

        with self.assertRaises(ValueError):
            sensor.moisture()

        # Make sure both pins are left in off state.
        self.assertFalse(pin_state[12])
        self.assertFalse(pin_state[16])
```

JeetShetty: Yeah I agree that we should be testing that the pins are left in the "off" state. I like your solution! Minus the large amount of code. Can you reiterate for me why you don't like using ```setUp```? I think the idea is that someone might only be interested in reading the code for some tests, and not others, and it makes it more difficult to have to scroll to ```setUp``` to figure out what's going on? In a case like this, it seems to me like it's a good way to share code because otherwise we're kind of clogging the tests anyway, and if we only have two tests it isn't difficult to get to the top of the file and look at ```setUp```. 

In the meantime I'll try to think of other ways to do this more concisely.

mtlynch: Oh yeah! We can do it in `setUp`. I was thinking about trying to extract just the IO stuff to its own separate helper class and didn't think about `setUp` but you're right. This should work:

```python
    def setUp(self):
        self.mock_adc = mock.Mock()
        self.mock_adc.read_adc.side_effect = ValueError()
        self.pin_state = collections.defaultdict(lambda: False)

        def mock_turn_pin_on(pin):
            self.pin_state[pin] = True

        def mock_turn_pin_off(pin):
            self.pin_state[pin] = False

        self.mock_pi_io = mock.Mock()
        self.mock_pi_io.turn_pin_on.side_effect = mock_turn_pin_on
        self.mock_pi_io.turn_pin_off.side_effect = mock_turn_pin_off
        self.mock_clock = mock.Mock()
```

>Can you reiterate for me why you don't like using setUp? I think the idea is that someone might only be interested in reading the code for some tests, and not others, and it makes it more difficult to have to scroll to setUp to figure out what's going on? 

Oh, I don't dislike `setUp` in general. I just think we need to be deliberate about what we put in it. Like [`test_poller.py`](https://github.com/JeetShetty/GreenPiThumb/blob/master/tests/test_poller.py) I think uses `setUp` well. But if we changed it to be like:

```python
    def setUp(self):
        self.clock_wait_event = threading.Event()
        self.mock_local_clock = mock.Mock()
        self.mock_sensor = mock.Mock()
        self.mock_store = mock.Mock()
        self.mock_sensor.temperature.return_value = 21.0 # <<< Bad

    def test_temperature_poller(self):
        temperature_poller = poller.TemperaturePoller(
            self.mock_local_clock, POLL_INTERVAL, self.mock_sensor,
            self.mock_store)
        self.mock_local_clock.now.return_value = TIMESTAMP_A
        self.mock_local_clock.wait.side_effect = (
            lambda _: self.clock_wait_event.set())

        temperature_poller.start_polling_async()
        self.clock_wait_event.wait(TEST_TIMEOUT_SECONDS)
        # BAD: Where is 21.0 coming from?
        self.mock_store.store_temperature.assert_called_with(TIMESTAMP_A, 21.0)
        self.mock_local_clock.wait.assert_called_with(POLL_INTERVAL)
```

Then it's confusing because it buries where the `21.0` is coming from so the reader has to search around.

`setUp` is good for building up scaffolding for the tests where the particulars don't matter much. It's bad for setting particular values that matter in the test.

mtlynch:
LGTM

Issue #49, opened 2017-01-08
Implement sleep window parser


mtlynch:
`pump_scheduler.py` relies on having a list of sleep windows where the pump doesn't run. We need to have some way for the user to actually communicate this.

I think the easiest would be a command line flag, like the user does:

```shell
python greenpithumb.py \
  --sleep_windows "12:30-1:30,22:00-6:00,8:00-8:01"
```

We could alternatively do it in a config file, but I find config files make things harder to just copy/paste easily.

JeetShetty:
I'm not sure how this works. How do we read in what the user is supplying?

mtlynch:
Oh, it's a command line flag. We're doing something really similar already with `--config_file` so when we run GreenPiThumb with

```shell
python greenpithumb.py \
  --config_file "C:\projects\greenpithumb\config.ini"
```

When we're in the `main` function in [greenpithumb.py](https://github.com/JeetShetty/GreenPiThumb/blob/master/greenpithumb/greenpithumb.py), `args.config_file` will be set to `"C:\projects\greenpithumb\config.ini"`.

So if we want to add a new flag, we can do it through `parser.add_argument` at the bottom of greenpithumb.py and then it will be available in the `args` parameter in `main` like `args.config_file`.

PR #56, opened 2017-01-18
Implement sleep window parser.

JeetShetty:
* Issue [#49](https://github.com/JeetShetty/GreenPiThumb/issues/49)

---
This change is [Reviewable](https://reviewable.io/reviews/jeetshetty/greenpithumb/56)


mtlynch:
Yep, looks good so far

tests/test_sleep_window_parser.py
```diff
@@ -0,0 +1,18 @@
...
+    def test_parser(self):
+        sleep_windows = '1:00-3:15,13:45-16:00,23:00-0:30'
+        parsed_windows = sleep_window_parser.parse(sleep_windows)
+        expected_parsed_windows = [
+            (datetime.time(1, 0), datetime.time(3, 15)),
+            (datetime.time(13, 45), datetime.time(16, 0)),
+            (datetime.time(23, 0), datetime.time(0, 30))
+        ]
+
+        self.assertEqual(parsed_windows, expected_parsed_windows)
```
mtlynch: Some other interesting test cases:

* `"banana"` (completely wrong input
* `"1:00-"` (incomplete range
* `"1-3"` (incorrectly formatted times: no colon)
* `"1:052-3:059"` (incorrectly formatted times: too many numbers)
* `"5:00-5:00"` (zero-sized range)
* `"5:00:09.897-5:00:15.394"` (unsupported seconds-level precision)
* `"01:00-01:05"` (leading zeroes in front of the hour)
* `"1:5-1:9"` (missing leading zeroes for minutes)

greenpithumb/sleep_window_parser.py
```diff
@@ -0,0 +1,26 @@
+"""Parses sleep window strings."""
```
mtlynch: Suggests naming this file `sleep_windows.py`.

`sleep_windows.parse` is clear and avoids the stutter-naming of "parser.parse", plus if we later decide to represent sleep windows differently, it'll be easy to add something here like `sleep_windows.Window`

greenpithumb/greenpithumb.py
```diff
@@ -68,4 +68,10 @@ def main(args):
         '--config_file',
         help='Wiring config file',
         default='wiring_config.ini')
+    parser.add_argument(
+        '-s',
+        '--sleep_windows',
+        type=str,
+        help='Sleep windows as 24-hour clock strings, with commas between each window',
```
mtlynch: I think we can add more detail here. We can make the help pretty long when needed like:

>Time windows during which GreenPiThumb will not activate its pump. The windows should be in the form of a comma-separated list of time ranges in 24-hour format, such as "22:00-8:00,3:15-3:45".

greenpithumb/sleep_window_parser.py
```diff
@@ -0,0 +1,26 @@
...
+
+import datetime
+
+
+def parse(sleep_windows_string):
+    """Converts sleep windows from strings to datetime objects.
+
+    Args:
+        sleep_windows_string: A string of sleep windows.
```
mtlynch: We should document here explicitly what the format has to be.

greenpithumb/sleep_windows.py
```diff
@@ -0,0 +1,84 @@
...
+    for minute in sleep_minutes:
+        awake_minutes[minute] = 0
+    if not sum(awake_minutes):
+        raise HypersomniaError(
+            'Sleep windows cannot cover the entire 24-hour range.')
+
+    return sleep_windows_datetime
+
+
+def _sleep_minutes(sleep_time, wake_time):
```
mtlynch: This feels like a pretty complicated way of making sure they don't specify 100% sleep time. I feel like if we have to do something complicated, that complicated thing might as well be coalescing overlapping/adjacent windows. I predict the coalesce logic will be simpler than building a set of every minute. Once we have coalesced windows, the check for hypersomnia is just checking `sleep_time == wake_time`.

The other advantage of coalescing is that it's independent of time granularity. If we at some point decide to allow seconds-level granularity, we have to change a lot of code in the current implementation, whereas with coalescing, it should just work no matter the time precision.

Alternatively, we could just not check for hypersomnia and so it's possible for the user to configure a pump that always sleeps.

greenpithumb/sleep_windows.py
```diff
@@ -0,0 +1,84 @@
...
+
+    Args:
+        sleep_time: Sleep time as a datetime object.
+        wake_time: Wake time as a datetime object.
+
+    Returns:
+        A set containing all of the minutes in the sleep window, from 0 to 1439.
+    """
+    sleep_minutes = set()
+    sleep_minute_no_hour = 60 * sleep_time.hour + sleep_time.minute
```
mtlynch: (maybe a moo point per my other comment, but): I'm not sure what's actually happening here. I can't figure out what the name `sleep_minute_no_hour` means. Oh, wait I just got it. I think a clearer name might be `total_minutes` because it mirrors what already exists in [`timedelta.total_seconds()`](https://docs.python.org/2/library/datetime.html#datetime.timedelta.total_seconds)

greenpithumb/sleep_windows.py
```diff
@@ -0,0 +1,84 @@
...
+
+def _sleep_minutes(sleep_time, wake_time):
+    """Calculates the individual minutes in given sleep window.
+
+    Args:
+        sleep_time: Sleep time as a datetime object.
+        wake_time: Wake time as a datetime object.
+
+    Returns:
+        A set containing all of the minutes in the sleep window, from 0 to 1439.
```
mtlynch: (maybe a moo point per my other comment, but): This could use a bit more detail. Saying "from 0 to 1439" gives the impression that the return set always includes all of those integers.

Also the number `1439` is confusing because it's not immediately recognizable as 24 * 60 - 1, so for things like that, suggest including the calculation too so it doesn't seem like a magic number.

greenpithumb/sleep_windows.py
```diff
@@ -0,0 +1,84 @@
...
+    """
+    sleep_minutes = set()
+    sleep_minute_no_hour = 60 * sleep_time.hour + sleep_time.minute
+    wake_minute_no_hour = 60 * wake_time.hour + wake_time.minute
+    if wake_time < sleep_time:
+        sleep_minutes.update(range(sleep_minute_no_hour, 1440))
```
mtlynch: Would rewrite the `1440` as `24 * 60` to make it obvious it's minutes per day. Or maybe even a module-level constant like `_MINUTES_PER_DAY = 24 * 60`

greenpithumb/sleep_windows.py
```diff
@@ -0,0 +1,84 @@
...
+                'Time windows must be in string format.')
+        if not window_match:
+            raise InvalidWindowFormatError(
+                'Sleep windows must be in 24-hour HH:MM-HH:MM format')
+        try:
+            sleep_time = datetime.time(
+                int(window_match.group(1)), int(window_match.group(2)))
+            wake_time = datetime.time(
+                int(window_match.group(3)), int(window_match.group(4)))
+        except (ValueError, TypeError):
```
mtlynch: Coveralls says we're never hitting this. I think because our invalid hour/minute cases are *too* off so the regex filters them. Can we add tests where it's like `'23:15-24:00'` and `'22:56-22:60'`?

tests/test_sleep_windows.py
```diff
@@ -0,0 +1,48 @@
...
+            ['100:50-03:00'],  # invalid hour
+            ['10:500-12:00'],  # invalid minute
+            ['05:00:09.897-06:00']  # unsupported seconds-level precision
+        ]
+        for invalid_input in invalid_inputs:
+            with self.assertRaises(sleep_windows.InvalidWindowFormatError):
+                sleep_windows.parse(invalid_input)
+
+    def test_hypersomnia(self):
+        hypersomnia_inputs = [['00:00-00:00'], ['01:00-12:00', '12:00-01:00'],
```
mtlynch: (minor) Suggest using parens for the outer set just to make the distinction a little clearer like

```python
hypersomnia_inputs = (['00:00-00:00'], ...
```

greenpithumb/sleep_windows.py
JeetShetty: Ah, yeah I agree that coalescing the sleep windows is a better option. I ended up deciding to not check for hypersomnia because it seems like a very low priority feature but adds a decent amount of logic to our code.

tests/test_sleep_windows.py
```diff
@@ -0,0 +1,48 @@
...
+        # Subset ranges OK.
+        self.assertEqual(
+            sleep_windows.parse(['01:00-03:15', '02:45-03:00']),
+            [(datetime.time(1, 0), datetime.time(3, 15)),
+             (datetime.time(2, 45), datetime.time(3, 0))])
+
+    def test_parser_invalid_inputs(self):
+        invalid_inputs = [
+            [''],  # empty string
+            [10],  # not a string
```
JeetShetty: Oh, good to know!

greenpithumb/greenpithumb.py
```diff
@@ -70,12 +70,11 @@ def main(args):
...
+        '--sleep_window',
         action='append',
         type=str,
-        help=('Time windows during which GreenPiThumb will not activate its' +
-              'pump. Windows should be in the form of a comma-separated list' +
-              'of time ranges in 24-hour format, such as' +
-              '"22:00-08:00,03:15-03:45"'),
-        default=None)
+        help=('Time windows during which GreenPiThumb will not activate its '
+              'pump. Windows should be in the form of a comma-separated list '
```
mtlynch: Isn't this no longer true?

PR #80, opened 2017-03-12
Adding RecordProcessor class

mtlynch:
Adds a RecordProcessor class that monitors the records queue and places the
records that arrive through the queue into the appropriate database store.

### DIFF
--- a/greenpithumb/greenpithumb.py
+++ b/greenpithumb/greenpithumb.py
@@ -1,6 +1,6 @@
 import argparse
 import contextlib
-import time
+import Queue
 
 import Adafruit_DHT
 import Adafruit_MCP3008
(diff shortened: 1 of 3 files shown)


greenpithumb/greenpithumb.py
```diff
@@ -47,35 +49,29 @@ def __init__(self, wiring_config):
...
-        print '%s\t%s\t\t%s\t\t%s' % (light_reading, moisture_reading,
-                                      temperature_reading, humidity_reading)
-
 
 def read_wiring_config(config_filename):
     with open(config_filename) as config_file:
         return wiring_config_parser.parse(config_file.read())
 
 
+def create_record_processor(db_connection, record_queue):
```
mtlynch: Yeah, I've thought about it, but I couldn't think of a good reason to underscore them. It's sort of like in tests where it just feels like extra keystrokes for no value.

PR #88, opened 2017-03-23
Adding picamera to requirements

mtlynch:


mtlynch:
Whoops, remembered that it breaks Travis

PR #92, opened 2017-03-24
Fix PumpScheduler to handle updated sleep window format.

JeetShetty:


greenpithumb/pump_scheduler.py
```diff
@@ -1,27 +1,31 @@
...
+            local_clock: A local clock interface
+            sleep_windows: A list of 2-tuples, each representing a sleep window.
+                Tuple items are datetime.time objects.
         """
-        self._clock = clock
+        self._local_clock = local_clock
         self._sleep_windows = sleep_windows
 
     def is_running_pump_allowed(self):
         """Returns True if OK to run pump, otherwise False."""
```
mtlynch: We should define explicitly what makes it OK to run the pump. In particular, we should say that it's not allowed from sleep time (inclusive) until wake time (exclusive).

greenpithumb/pump_scheduler.py
```diff
@@ -1,27 +1,31 @@
...
-        current_hour = self._clock.now().hour
+        now = self._local_clock.now()
+        current_time = datetime.time(now.hour, now.minute)
```
mtlynch: We can simplify these two lines to `current_time = self._local_clock.now().time()`

greenpithumb/pump_scheduler.py
```diff
@@ -1,27 +1,31 @@
...
 
-        for sleep_hour, wake_hour in self._sleep_windows:
-            if wake_hour < sleep_hour:
-                if current_hour >= sleep_hour or current_hour < wake_hour:
+        for sleep_time, wake_time in self._sleep_windows:
+            if wake_time < sleep_time:
```
mtlynch: Suggest adding a comment here like `# Check if sleep interval wraps overnight` because it's a little bit subtle.

PR #93, opened 2017-03-25
Rearchitecting database connection

mtlynch:
Rearchitecting the database strategy to get rid of the RecordProcessor class.
I couldn't figure out a way to do thread-safe DB reads under that system.

Instead, each poller thread has its own DB connection that the DB store owns.
We're still instantiating the DB connection in the main thread and using it in
worker threads, but that's okay as long as multiple threads aren't operating
on the same connection instance. We're turning off sqlite's thread-safety
checks because it will raise exceptions for creating in one thread and using
in another even though that's actually okay.

This basically rolls back a lot of changes from 69f92f, but it keeps records
as public objects that go into or come out of the DbStore classes.


mtlynch:
Wait, nevermind. I think I have a better solution for this.

PR #96, opened 2017-03-26
Refactoring pollers

mtlynch:
The goal of the refactoring is to fix two problems:
1. SensorPollBase has a design flaw in that it runs in the main thread, but
 holds instance data that belongs to in the background thread.
 This is error-prone because if any of the member variables are not thread-safe
 we it creates concurrency issues if both threads touch the members at the same
 time. We should redesign it so that the variables that the background thread uses
 are separate from the variables that the main thread and kept in private members
 so they're inaccessible outside the thread.
2. There are two ways of instantiating the pollers: through the
 SensorPollerFactory and by instantiating the objects directly. Offering an
 API with multiple ways of doing the same thing is messy, so we're making
 the classes private so that the factory is the only way clients can create
 them. We also rewrite the unit tests to test through the factory so that the
 unit tests are an accurate reflection of how the caller uses the APIs.

### DIFF
--- a/greenpithumb/poller.py
+++ b/greenpithumb/poller.py
@@ -7,165 +7,138 @@ logger = logging.getLogger(__name__)
 
 
 class SensorPollerFactory(object):
-    """Factory to simplify the semantics of creating pollers."""
+    """Factory for creating sensor poller objects."""
 
     def __init__(self, local_clock, poll_interval, record_queue):
+        """Create a new SensorPollerFactory instance.
+
+        Args:
+            local_clock: A local time zone clock interface.
+            poll_interval: An int of how often the data should be polled for,
+                in seconds.
+            record_queue: Queue on which to place database records.
+        """
         self._local_clock = local_clock
         self._poll_interval = poll_interval
         self._record_queue = record_queue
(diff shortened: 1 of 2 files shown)


mtlynch:
>Can you expand on point 1? I'm not quite following for some reason.

Oh yeah, I didn't explain it very well. I updated the PR description to try to clarify.

To add context, in one of my attempts at refactoring, I added a database connection object, so it looked kind of like:

```python
 class SensorPollerBase(object):
     """Base class for sensor polling."""
 
     def __init__(self, local_clock, poll_interval):
         self._local_clock = local_clock
         self._poll_interval = poll_interval
         self._db_connection = self._init_db_connection()
         self._closed = threading.Event()
```

And then the background thread would use the `self._db_connection` for writing to the DB. But then the problem was I did this:

```python
def close(self):
  self._db_connection.close()
  self._closed.set()
```

But that blew up because `close()` gets called in the main thread, but the background thread actually owns `self._db_connection` and it's not thread safe. So I realized it doesn't make sense for all of the variables to be members of the same object if they're owned by different threads.

PR #97, opened 2017-03-26
Refactoring DbStore classes

mtlynch:
Refactoring the DbStore classes to reduce the redundant code between classes.

### DIFF
--- a/greenpithumb/db_store.py
+++ b/greenpithumb/db_store.py
@@ -3,7 +3,7 @@ import logging
 import os
 import sqlite3
 
-from dateutil import parser
+from dateutil import parser as date_parser
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


greenpithumb/db_store.py
```diff
@@ -110,20 +110,48 @@ def open_or_create_db(db_path):
...
-        """Creates a new DbStoreBase object for storing information.
+        """Creates a new _DbStoreBase object for storing information.
 
         Args:
             connection: SQLite database connection.
         """
         self._connection = connection
         self._cursor = connection.cursor()
 
+    def _do_insert(self, sql, timestamp, value):
```
JeetShetty: Why ```_do_insert``` instead of ```_insert```? Ditto for ```_do_get```

mtlynch: It's a pattern that I've seen a bit to try to make it obvious which function is doing the real heavy lifting. Plus, I'm nervous about having `insert` and `_insert` because I feel like it's too easy to misread/mistype one for the other.

Issue #99, opened 2017-03-27
Can't kill GreenPiThumb with Ctrl+C


mtlynch:
Because of something weird we're doing with threading, GreenPiThumb doesn't exit when the user performs a `KeyboardInterrupt` (hits Ctrl+C). It eventually quits on the next poll, but the user can't quit immediately.

I think it's because the main thread is stuck blocking on reading from the record processing queue and can't be interrupted until something comes through.

PR #100, opened 2017-03-27
Changing polling to occur on time boundaries

mtlynch:
In the current implementation, each sensor starts reading immediately, then
polls every n seconds. This means that the readings depend on when the
greenpithumb started, so if a user runs greenpithumb on two different days
with two different start times, the reading timestamps won't be aligned.

By changing the polling time to happen on interval boundaries, we get aligned
readings no matter when we start the greenpithumb process.

As a side effect, this doesn't solve the uninterruptible process issue (#99)
but it should make it easier to fix since we're getting rid of most of the
threads that block forever until some event happens.

### DIFF
--- a/greenpithumb/poller.py
+++ b/greenpithumb/poller.py
@@ -60,7 +60,7 @@ class SensorPollerFactory(object):
 def _datetime_to_unix_time(dt):
     """Converts a datetime into seconds since UNIX epoch."""
     unix_epoch = datetime.datetime(year=1970, month=1, day=1, tzinfo=pytz.utc)
-    return (dt - unix_epoch).total_seconds()
+    return int((dt - unix_epoch).total_seconds())
 
 
 def _round_up_to_multiple(value, multiple):
@@ -128,7 +128,7 @@ class _SensorPollWorkerBase(object):
         """
         next_poll_time = _round_up_to_multiple(self._unix_now(),
                                                self._poll_interval)
-        if next_poll_time and (next_poll_time == last_poll_time):
+        if last_poll_time and (next_poll_time == last_poll_time):
             next_poll_time += self._poll_interval
         return next_poll_time
 
--- a/greenpithumb/poller.py
+++ b/greenpithumb/poller.py
@@ -1,9 +1,17 @@
+import datetime
 import logging
 import threading
+
+import pytz
 
 import db_store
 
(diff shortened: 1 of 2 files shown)


greenpithumb/poller.py
```diff
@@ -73,12 +97,54 @@ def __init__(self, local_clock, poll_interval, record_queue, sensor):
...
+        Args:
+            last_poll_time: The UNIX time of the last poll or None if this is
+                the first poll.
+
+        Returns:
+            UNIX time of next scheduled poll.
+        """
+        next_poll_time = _round_up_to_multiple(self._unix_now(),
+                                               self._poll_interval)
+        if next_poll_time and (next_poll_time == last_poll_time):
```
JeetShetty: I'm confused here. ```next_poll_time``` will never be False, right? Also ```next_poll_time``` and ```last_poll_time``` will always be different since ```total_seconds()``` returns to microsecond precision?

mtlynch: >I'm confused here. next_poll_time will never be False, right?

Oh, whoops, that was supposed to be `last_poll_time`. Fixed.

>Also next_poll_time and last_poll_time will always be different since total_seconds() returns to microsecond precision?

Oh yeah, you're right. I didn't realize `total_seconds` was a float instead of an int. It *theoretically* could end up the same value if it managed to poll in less than 1 microsecond (which probably can't happen practically), but I actually meant for the UNIX time to be integer seconds because I feel like it's kind of confusing / unexpected to see mod operations on a float. So I've updated `_unix_now` to return an int and now the check is more necessary because there can be a time collision if the sensor reading takes less than one second.

mtlynch:
>It seems to me like we only need to sync the pollers before the first poll, and then we can just stick to what we had before where we poll every poll_interval seconds. But, it seems like you made a deliberate choice to do it this way, so let me know if there's a good reason for it!

Doesn't that assume that sensor readings take 0 time, though? Because if we alight to a boundary and then wait a fixed amount between readings, I think we'll end up skewing due to the time of reading a sensor.

But the primary driver is that the background threads need to quickly notice if the main thread has exited so they can shut down gracefully. In the previous implementation, where they just called `clock.wait` (which in turn calls `time.sleep`) they're blocked and don't notice if anyone callls `stop`.

PR #104, opened 2017-03-31
Changes main script so user can Ctrl+C to exit

### LINKED ISSUE #99: Can't kill GreenPiThumb with Ctrl+C

mtlynch:
Changes the main script and record processor so that we're not blocking
indefinitely until a polling event occurs.

This fixes #99

### DIFF
--- a/greenpithumb/greenpithumb.py
+++ b/greenpithumb/greenpithumb.py
@@ -116,7 +116,7 @@ def main(args):
                 current_poller.start_polling_async()
             while True:
                 record_processor.try_process_next_record()
-                time.sleep(0.5)
+                time.sleep(0.1)
         except KeyboardInterrupt:
             logger.info('Caught keyboard interrupt. Exiting.')
         finally:
--- a/greenpithumb/greenpithumb.py
+++ b/greenpithumb/greenpithumb.py
@@ -2,6 +2,7 @@ import argparse
 import contextlib
 import logging
 import Queue
+import time
 
 import Adafruit_DHT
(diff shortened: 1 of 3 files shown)


greenpithumb/greenpithumb.py
```diff
@@ -114,7 +115,10 @@ def main(args):
             for current_poller in pollers:
                 current_poller.start_polling_async()
             while True:
-                record_processor.process_next_record()
+                record_processor.try_process_next_record()
+                time.sleep(0.5)
```
JeetShetty: Why are we sleeping here?

mtlynch: Otherwise it spikes the CPU because the program has nothing to do except keep calling `try_process_next_record` nonstop. With `sleep` the OS knows the program has nothing to do, so it doesn't use the CPU in that thread.

mtlynch: It's a little too long though because 0.5 means that if 5 events happen at once, it'll take us 2.5 seconds to process. It probably would make sense to have `try_process_next_record` return `True` or `False` depending on if it processed anything and we only sleep when it's empty.

For now I'll lower to 0.1.

Issue #105, opened 2017-03-31
Race condition in poller tests


mtlynch:
I'm seeing occasional failures in Travis:

```
======================================================================
ERROR: test_ambient_light_poller (tests.test_poller.SimplePollerClassesTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/JeetShetty/GreenPiThumb/tests/test_poller.py", line 94, in test_ambient_light_poller
    self.record_queue.get(block=True, timeout=TEST_TIMEOUT_SECONDS))
  File "/opt/python/2.7.9/lib/python2.7/Queue.py", line 176, in get
    raise Empty
Empty

----------------------------------------------------------------------
Ran 79 tests in 1.596s

FAILED (errors=1)
```

Looks like there's a timing issue that causes the tests to sometimes fail.

mtlynch:
I think #136 fixed this

PR #107, opened 2017-03-31
Increase seconds till time out in poller tests.

JeetShetty:
#105 

I'm not sure why, but when I change the timeout seconds from 1.5 back to 3.0, this problem goes away for me. 1.5 seconds seems like it should be plenty. Do you know what's going on?


mtlynch:
>Do you know what's going on?

Not sure, but this is a decent workaround until we can figure out the root cause.

PR #116, opened 2017-04-01
Adding poller for camera

mtlynch:
Adding a separate photo interval flag to control frequency of photos and log output for the camera manager.

### DIFF
--- a/greenpithumb/greenpithumb.py
+++ b/greenpithumb/greenpithumb.py
@@ -6,9 +6,11 @@ import time
 
 import Adafruit_DHT
 import Adafruit_MCP3008
+import picamera
 import RPi.GPIO as GPIO
 
 import adc_thread_safe
+import camera_manager
 import clock
 import db_store
 import dht11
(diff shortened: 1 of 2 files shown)


greenpithumb/greenpithumb.py
```diff
@@ -72,6 +74,15 @@ def pump_if_needed(self, _):
     ]
 
 
+def make_camera_poller(photo_interval, image_path, record_queue):
+    local_clock = clock.LocalClock()
+    poller_factory = poller.SensorPollerFactory(local_clock, photo_interval,
+                                                record_queue)
```
JeetShetty: (Optional) Feels strange to create a new poller factory just for the camera poller. My first thought was to just add a ```poll_interval``` argument to ```create_camera_poller```, but that also doesn't feel good.

mtlynch: Yeah, I agree but I couldn't think of anything either. My initial approach was to add a photo_interval parameter to the main sensor factory but that felt klunkier.

PR #119, opened 2017-04-03
Changing camera to save hi res files and use UTC time

mtlynch:
Changing camera poller to use UTC time and a string timestamp format to match
the database timestamp format.

Changing the camera's resolution to be the max resolution.

### DIFF
--- a/greenpithumb/camera_manager.py
+++ b/greenpithumb/camera_manager.py
@@ -3,31 +3,34 @@ import os
 
 logger = logging.getLogger(__name__)
 
+# Format of filename to write for camera image file (assumes timestamp is in
+# UTC), as YYYY-MM-DDTHH:MM:SSZ (ISO 8601 format)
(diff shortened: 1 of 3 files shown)


greenpithumb/camera_manager.py
```diff
@@ -3,31 +3,34 @@
 
 logger = logging.getLogger(__name__)
 
+# Format of filename to write for camera image file (assumes timestamp is in
+# UTC), as YYYY-MM-DDTHH:MM:SSZ (ISO 8601 format)
+_FILENAME_FORMAT = '%Y-%m-%dT%H%M%SZ.jpg'
```
JeetShetty: (optional) Seconds level precision also seems unnecessary.

mtlynch: Oh good point. The encoder in the frontend will still JSONify it with seconds-level precision, but maybe we can address that later.

PR #122, opened 2017-04-05
Getting rid of seconds-level precision

### LINKED ISSUE #121: Drop seconds-level granularity
Seconds-level granularity doesn't really make sense for anything in GreenPiThumb. The user doesn't need to schedule sensor polling intervals at the level of seconds and it doesn't make sense to take a picture of a plant at a frequency > 1x per min.

We should:
* Change command-line flags to take minutes rather than seconds
* Change the database to drop the seconds from the serialized timestamps

mtlynch:
Now the smallest level of granularity in GreenPiThumb is minutes. Polling
happens in intervals with minute granularity and timestamps are saved with
minutes-level precision.

This fixes #121.

### DIFF
--- a/tests/test_db_store.py
+++ b/tests/test_db_store.py
@@ -45,15 +45,15 @@ class OpenOrCreateTest(unittest.TestCase):
             cursor = connection.cursor()
             # Insertions into all tables should work after initialization.
             cursor.execute('INSERT INTO temperature VALUES (?, ?)',
-                           ('2016-07-23T10:51:09Z', 98.6))
+                           ('2016-07-23T10:51Z', 98.6))
             cursor.execute('INSERT INTO ambient_humidity VALUES (?, ?)',
-                           ('2016-07-23T10:51:09Z', 93.7))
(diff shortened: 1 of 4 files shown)


tests/test_db_store.py
```diff
@@ -68,45 +68,45 @@ def test_insert_soil_moisture(self):
         """Should insert timestamp and moisture level into database."""
         record = db_store.SoilMoistureRecord(
             timestamp=datetime.datetime(
-                2016, 7, 23, 10, 51, 9, tzinfo=pytz.utc),
+                2016, 7, 23, 10, 51, tzinfo=pytz.utc),
```
JeetShetty: Oh I think the insert tests should still have timestamps that go down to microsecond precision since that's what ```clock.now()``` returns.

JeetShetty: Oh I think the test insert timestamps should still go down to millisecond precision since that's how they get passed in from ```poller```.

mtlynch: Oh, good point. Fixed.

Issue #117, opened 2017-04-01
Camera poller needs to close camera manager


mtlynch:
Camera Manager needs to be closed when it is no longer needed, but the poller is not closing it.

PR #127, opened 2017-04-07
Update camera poller to close the camera when polling is complete.

JeetShetty:
#117


greenpithumb/poller.py
```diff
@@ -230,6 +230,11 @@ def _poll_once(self):
         """Captures and stores an image."""
         self._sensor.save_photo()
 
+    def stop(self):
+        """End worker polling and close camera."""
+        self._stopped.set()
+        self._sensor.close()
```
mtlynch: Suggest:

```python
        self._sensor.close()
        super(_CameraPollWorker, self).stop()
```

Because then if we ever add anything to the base class' `close`, we don't have to remember to add it here independently. Also I'm suggesting calling the parent's `close` second because it's the opposite of a constructor. The constructor we want to call the parent first because the child might have dependencies on the parent's object initialization. So when we close, we want to close the child's things first, then close the parent's in case the child has any dependencies on the parent's objects staying alive.

JeetShetty: Oh yeah! Good points.

Issue #83, opened 2017-03-17
pi_io needs to be thread-safe


mtlynch:
It's shared by two different threads, so it needs sane behavior when multiple threads try to close it and when threads try to call its normal methods after another thread has closed it.

JeetShetty:
Shouldn't the main thread be the only calling ```close```?

mtlynch:
>Shouldn't the main thread be the only one calling close?

Yeah that's true.

PR #128, opened 2017-04-07
Update main thread to close Pi IO.

JeetShetty:
#83


greenpithumb/greenpithumb.py
```diff
@@ -29,7 +29,7 @@
 
 
 def make_sensor_pollers(poll_interval, wiring_config, record_queue,
-                        sleep_windows):
+                        sleep_windows, raspberry_pi_io):
```
mtlynch: Haha stupid Python name precedence rules!

Issue #131, opened 2017-04-08
Pre-commit hook doesn't block on YAPF diff


mtlynch:
The pre-commit hook should block the commit if the YAPF diff turns up a delta, but currently it seems to just print a message and keep going. We should change the `build` so that it returns a failing exit code on a non-zero diff.

mtlynch:
Apparently this is a behavior change in YAPF:

https://github.com/google/yapf/issues/325

They're considering a fix, but there's a workaround possible:

https://github.com/google/yapf/issues/325#issuecomment-261552890

PR #135, opened 2017-04-09
Fixing build script to return non-zero exit on YAPF diff

mtlynch:
YAPF changed behavior so it no longer returns a non-zero exit code on a diff
(see https://github.com/google/yapf/issues/325). This updates the build script
so that the build fails if there's a YAPF diff.


mtlynch:
This fixes #131.

PR #137, opened 2017-04-13
Fixing last_poll_time function

mtlynch:
last_poll_time was returning timestamps as ints, whereas the callers were
expecting datetimes. Also, the way that last_poll_time was getting updated
wasn't really intuitive, so updated the code so that last_poll_time only
updates after the scheduler waits through a poll boundary.

### DIFF
--- a/greenpithumb/poller.py
+++ b/greenpithumb/poller.py
@@ -102,7 +102,7 @@ class Scheduler(object):
             self._unix_now(), int(self._poll_interval.total_seconds()))
         if self._last_poll_time and (next_poll_time == self._last_poll_time):
             next_poll_time += int(self._poll_interval.total_seconds())
-        self._last_poll_time = next_poll_time
+
         return next_poll_time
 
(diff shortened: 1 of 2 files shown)


greenpithumb/poller.py
```diff
@@ -102,7 +102,7 @@ def _next_poll_time(self):
             self._unix_now(), int(self._poll_interval.total_seconds()))
         if self._last_poll_time and (next_poll_time == self._last_poll_time):
```
JeetShetty: Doesn't this line not make sense now since ```next_poll_time``` is a POSIX timestamp and ```_last_poll_time``` is a datetime object?

mtlynch:
>When reading the code I got confused by the fact that last_poll_time() returns a different type than _last_poll_time.

Yeah, that's true. It felt a little weird as I was doing it and now that you mention it, I think we can do a better job of distinguishing `datetime`s from UNIX timestamps. Updated to clarify.

Issue #118, opened 2017-04-01
Camera manager should not take pictures at night


mtlynch:
We should integrate with the light sensor to detect when it's too dark to take a photo

JeetShetty:
Can I just add the light sensor to the ```_CameraPollWorker``` class? Are there any issues with the light sensor being read too frequently?

mtlynch:
Yeah that should work. I don't think there's much of a cost to reading the light sensor a lot.

PR #139, opened 2017-04-15
Update camera poller to only take pictures in sufficient light.

JeetShetty:
Until now I wasn't really thinking about how the camera poller has access to the queue but doesn't use it. Seems bad? Should I try to find a way to refactor the camera poller code?

#118


greenpithumb/camera_manager.py
```diff
@@ -6,6 +6,7 @@
 # Format of filename to write for camera image file (assumes timestamp is in
 # UTC), as YYYY-MM-DDTHH:MMZ (minutes-level precision).
 _FILENAME_FORMAT = '%Y-%m-%dT%H%MZ.jpg'
+LIGHT_THRESHOLD_PCT = 75
```
mtlynch: Also, can we add a comment explaining what this is? Also, I think 60% is enough for decent pictures.

greenpithumb/poller.py
```diff
@@ -256,9 +257,23 @@ def _poll_once(self):
...
+            camera_manager: Manages the capture and saving of photos.
+            light_sensor: An interface for reading the light level.
+        """
+        super(_CameraPollWorker, self).__init__(scheduler, record_queue,
+                                                camera_manager)
+        self._light_sensor = light_sensor
+
     def _poll_once(self):
         """Captures and stores an image."""
-        self._sensor.save_photo()
```
mtlynch: Can we push more of the logic into `CameraManager`? Sort of mirroring the design of `PumpManager` with its moisture threshold and `pump_if_needed`.

mtlynch:
>Until now I wasn't really thinking about how the camera poller has access to the queue but doesn't use it. Seems bad? Should I try to find a way to refactor the camera poller code?

Oh yeah, it's a little messy, but I can't think of a way to fix it without adding more complexity. The way that first comes to mind is to create a deeper class hierarchy like

```python
class _SensorPollWorkerBase(object):
  # Base for all poll workers

class RecordProducingPollWorkerBase(_SensorPollWorkerBase):
  # Base class for workers that produce recorsd

class TemperatureWorker(RecordProducingPollWorkerBase):
  ...
```

But I feel like that's more confusing than what we have now.

Any ideas for different approaches? I think as-is it's fine, but maybe there's a clean way to fix it.

JeetShetty:
>Any ideas for different approaches? I think as-is it's fine, but maybe there's a clean way to fix it.

Yeah nothing good comes to mind right now, but I'll keep thinking about it. In the most recent commit I changed the ```camera_poller_factory``` instantiation to use ```None``` instead of the actual record queue, but I'm not sure if this is a good idea.

PR #143, opened 2017-05-12
Add a max interval between waterings.

JeetShetty:


greenpithumb/pump.py
```diff
@@ -123,3 +129,16 @@ def is_running_pump_allowed(self):
                     return False
 
         return True
+
+    def pump_interval_exceeded(self):
+        """Returns True if plant needs to be watered now, False otherwise.
+
+        Plant needs to be watered regardless of soil moisture level if the time
+        elapsed since the last watering is greater than the pump interval.
+        """
```
mtlynch: Suggest describing it in terms of the pump rather than the plant since the pump's behavior is independent of the plant.

greenpithumb/pump.py
```diff
@@ -83,9 +83,11 @@ def pump_if_needed(self, moisture):
             The amount of water pumped, in mL.
         """
         if self._pump_scheduler.is_running_pump_allowed():
-            if moisture < self._moisture_threshold:
+            if moisture < self._moisture_threshold or (
+                    self._pump_scheduler.pump_interval_exceeded()):
```
mtlynch: (optional): The logic's getting a bit complex here. Consider splitting out something like

```python
def _should_pump(self, moisture):
  if not self._pump_schedulre.is_running_pump_allowed():
    return False
  return (moisture < self._moisture_threshold) or self._pump_scheduler.pump_interval_exceeded()
```

greenpithumb/pump.py
```diff
@@ -94,16 +96,20 @@ def pump_if_needed(self, moisture):
 class PumpScheduler(object):
     """Controls when the pump is allowed to run."""
 
-    def __init__(self, local_clock, sleep_windows):
+    def __init__(self, local_clock, sleep_windows, pump_interval):
```
mtlynch: This feels a bit weird because it's outside the scope of what `PumpScheduler` is defined to do. But also it feels like it fits in better with `PumpManager` because then `PumpManager` is in charge of turning the pump on, `PumpScheduler` is responsible for keeping the pump off.

I think there's a temptation to go `PumpScheduler` because it knows what time it is, but what if we created something like a `Timer` class like that we could call `timer.done()` and `timer.reset()` and gave that to the `PumpManager` so it still only needs one new param?

tests/test_clock.py
```diff
@@ -30,3 +32,57 @@ def test_negative_numbers_raise_ValueError(self):
...
+
+    def test_timer_duration_exceeded(self):
+        duration = datetime.timedelta(hours=3 * 24)
+        self.mock_clock.now.side_effect = [
+            datetime.datetime(
+                2016, 7, 23, 10, 51, 9, 928000, tzinfo=pytz.utc),
+            datetime.datetime(
+                2016, 7, 27, 10, 51, 9, 928000, tzinfo=pytz.utc)
+        ]
+        self.assertTrue(clock.Timer(self.mock_clock, duration).expired())
```
mtlynch: (minor) Consider using rounder numbers to make it easier to see at a glance what the difference is between the two dates like

```python
            datetime.datetime(
                2000, 1, 1, 0, 0, 0, 0, tzinfo=pytz.utc),
            datetime.datetime(
                2000, 1, 4, 0, 0, 0, 0, tzinfo=pytz.utc)
```

tests/test_clock.py
```diff
@@ -30,3 +32,57 @@ def test_negative_numbers_raise_ValueError(self):
...
+
+    def test_timer_duration_matched(self):
+        duration = datetime.timedelta(hours=3 * 24)
+        self.mock_clock.now.side_effect = [
+            datetime.datetime(
+                2016, 7, 23, 10, 51, 9, 928000, tzinfo=pytz.utc),
+            datetime.datetime(
+                2016, 7, 26, 10, 51, 9, 928000, tzinfo=pytz.utc)
+        ]
+        self.assertFalse(clock.Timer(self.mock_clock, duration).expired())
```
mtlynch: I expected this  to be `True`. If we set a 5 minute timer at 1pm, wouldn't we want it to expire at 1:05pm rather than 1:05:00.000001pm?

tests/test_clock.py
```diff
@@ -30,3 +32,57 @@ def test_negative_numbers_raise_ValueError(self):
...
+        timer = clock.Timer(self.mock_clock, None)
+        self.assertEqual(
+            timer._start_time,
+            datetime.datetime(
+                2016, 7, 23, 10, 51, 9, 928000, tzinfo=pytz.utc))
+        timer.reset()
+        self.assertEqual(
+            timer._start_time,
+            datetime.datetime(
+                2016, 7, 26, 10, 51, 9, 928000, tzinfo=pytz.utc))
```
mtlynch: Peeking into timer's internal variables is a bit ugly. Tests are generally more robust if they call the system under test the same way that a real client would. But this one is a bit awkward to test because it's hard to verify from the outside that expire actually reset the clock correctly.

I feel like it's sufficient to do something like:

1. Force timer to expiration
1. Reset
1. Assert that timer is no longer expired
1. (maybe?) Verify we can expire it again

tests/test_clock.py
JeetShetty: You are correct!

tests/test_clock.py
JeetShetty: Ah yeah good point

tests/test_clock.py
```diff
@@ -43,46 +43,42 @@ def test_timer_duration_exceeded(self):
         duration = datetime.timedelta(hours=3 * 24)
         self.mock_clock.now.side_effect = [
             datetime.datetime(
-                2016, 7, 23, 10, 51, 9, 928000, tzinfo=pytz.utc),
-            datetime.datetime(
-                2016, 7, 27, 10, 51, 9, 928000, tzinfo=pytz.utc)
+                2000, 1, 1, 0, 0, 0, 0, tzinfo=pytz.utc), datetime.datetime(
+                    2000, 1, 5, 0, 0, 0, 0, tzinfo=pytz.utc)
         ]
```
mtlynch: Oh yeah, this is wonky. We can change the right bracket to:

```python
]  # yapf: disable
```

And then YAPF won't touch the line. Worth it when it gets it this bad. But maybe moot based on my comment below.

tests/test_clock.py
```diff
@@ -43,46 +43,42 @@ def test_timer_duration_exceeded(self):
...
         timer.reset()
-        self.assertEqual(
-            timer._start_time,
-            datetime.datetime(
-                2016, 7, 26, 10, 51, 9, 928000, tzinfo=pytz.utc))
\ No newline at end of file
+        # First expired call is one microsecond before when the timer should
+        # expire, second is right when it should expire.
+        self.assertFalse(timer.expired())
+        self.assertTrue(timer.expired())
```
mtlynch: I feel like by queueing up the `datetime.datetime` values, we're making it harder for the reader to understand the test. Because we start with all these times, but then it's on the reader to figure out how the times map to each line in the test, which is kind of hard.

Consider rewriting so that it's more obvious when the time changes are happening, e.g.

```python
    def test_reset(self):
        duration = datetime.timedelta(hours=3 * 24)
        self.mock_clock.now.return_value = datetime.datetime(
            2000, 1, 1, 0, 0, 0, 0, tzinfo=pytz.utc)
        timer = clock.Timer(self.mock_clock, duration)
        self.mock_clock.now.return_value = datetime.datetime(
            2000, 1, 2, 0, 0, 0, 0, tzinfo=pytz.utc)
        timer.reset()
        self.mock_clock.now.return_value = datetime.datetime(
            2000, 1, 4, 23, 59, 59, 999999, tzinfo=pytz.utc)
        # First expired call is one microsecond before when the timer should
        # expire, second is right when it should expire.
        self.assertFalse(timer.expired())
        self.mock_clock.now.return_value = datetime.datetime(
            2000, 1, 5, 0, 0, 0, 0, tzinfo=pytz.utc)
        self.assertTrue(timer.expired())
```

PR #146, opened 2017-05-27
Switching from PlantFriends soil sensor to SparkFun

mtlynch:
Changes code to use the SparkFun soil moisture sensor instead of the
PlantFriends soil moisture sensor. The SparkFun cannot reverse its polarity
like the PlantFriends, so it is constantly connected to ground and only needs
one GPIO pin to flip up and down for readings.

Updates the code and the Fritzing diagram.


greenpithumb/soil_moisture_sensor.py
```diff
@@ -15,43 +15,24 @@ def __init__(self, adc, pi_io, channel, gpio_pin_1, gpio_pin_2, clock):
             pi_io: Raspberry Pi I/O interface.
             channel: ADC channel to which the moisture sensor is connected. Must
                 be an int between 0 and 7.
-            gpio_pin_1: One of the Raspberry Pi GPIO pins that the moisture
-                sensor is connected to. Must be an int between 2 and 27.
-            gpio_pin_2: The other GPIO pin that the moisture sensor is connected
-                to. Must be an int between 2 and 27.
-            clock: Clock interface.
+            gpio_pin: The Raspberry Pi GPIO pins that the moisture sensor is
```
JeetShetty: 'pin' not 'pins' I think

mtlynch: Ah good catch! Fixed.

PR #150, opened 2017-06-16
Observe pump_interval across restarts

mtlynch:
Adds a lookup in the database on startup to find the last recorded time that
GreenPiThumb ran the pump.

To support this, changes the Timer interface to support setting an explicit
time remaining that may differ from the duration.

### DIFF
--- a/greenpithumb/clock.py
+++ b/greenpithumb/clock.py
@@ -46,6 +46,17 @@ class Timer(object):
         self.reset()
 
     def set_remaining(self, time_remaining):
+        """Adjusts the remaining time on the timer.
+
+        Args:
+            time_remaining: The new amount of time remaining before the timer
+                expires. Must be a timedelta between 0 seconds and the timer
+                duration (inclusive).
+
+        Raises:
+            ValueError if time_remaining is a negative timedelta or is longer
+            than the timer's duration.
+        """
         if time_remaining < datetime.timedelta(seconds=0):
             raise ValueError('time_remaining must be non-negative')
         if time_remaining > self._duration:
--- a/greenpithumb/clock.py
+++ b/greenpithumb/clock.py
@@ -43,12 +43,20 @@ class Timer(object):
         """
         self._clock = clock
         self._duration = duration
-        self._start_time = self._clock.now()
+        self.reset()
+
+    def set_remaining(self, time_remaining):
+        if time_remaining < datetime.timedelta(seconds=0):
+            raise ValueError('time_remaining must be non-negative')
+        if time_remaining > self._duration:
+            raise ValueError(
+                'Cannot set time_remaining to longer than duration')
+        self._end_time = self._clock.now() + time_remaining
 
     def expired(self):
         """Returns True if the countdown has expired."""
-        return (self._clock.now() - self._start_time) >= self._duration
+        return self._clock.now() >= self._end_time
 
     def reset(self):
         """Resets the countdown timer to its starting duration."""
-        self._start_time = self._clock.now()
+        self._end_time = self._clock.now() + self._duration
--- a/greenpithumb/greenpithumb.py
+++ b/greenpithumb/greenpithumb.py
@@ -150,8 +150,8 @@ def make_pump_manager(moisture_threshold, sleep_windows, raspberry_pi_io,
             (last_pump_time + pump_interval) - clock.Clock().now())
     else:
         logger.info('no previous watering found')
-        time_remaining = pump_interval
-    logger.info('max time until until next watering: %s', time_remaining)
+        time_remaining = datetime.timedelta(seconds=0)
+    logger.info('time until until next watering: %s', time_remaining)
     pump_timer.set_remaining(time_remaining)
     return pump.PumpManager(water_pump, pump_scheduler, moisture_threshold,
                             pump_amount, pump_timer)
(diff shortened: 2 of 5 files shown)


greenpithumb/greenpithumb.py
```diff
@@ -139,6 +141,13 @@ def make_pump_manager(moisture_threshold, sleep_windows, raspberry_pi_io,
                            clock.Clock(), wiring_config.gpio_pins.pump)
     pump_scheduler = pump.PumpScheduler(clock.LocalClock(), sleep_windows)
     pump_timer = clock.Timer(clock.Clock(), pump_interval)
+    last_pump_time = pump_history.last_pump_time(
+        db_store.WateringEventStore(db_connection))
+    logger.info('last watering was at %s', last_pump_time)
+    time_remaining = max(datetime.timedelta(seconds=0),
+                         (last_pump_time + pump_interval) - clock.Clock().now())
```
JeetShetty: I think this is going to raise an exception when the DB is empty since ```last_pump_time``` will be ```None```

greenpithumb/clock.py
```diff
@@ -43,12 +43,20 @@ def __init__(self, clock, duration):
         """
         self._clock = clock
         self._duration = duration
-        self._start_time = self._clock.now()
+        self.reset()
+
+    def set_remaining(self, time_remaining):
+        if time_remaining < datetime.timedelta(seconds=0):
+            raise ValueError('time_remaining must be non-negative')
```
JeetShetty: Wondering why you decided to add a check for this. I feel like in the past we've just put in the docstring what values are valid and left it at that.

tests/test_clock.py
```diff
@@ -69,6 +69,50 @@ def test_timer_duration_not_matched_or_exceeded(self):
...
+
+    def test_set_remaining_succeeds_when_remaining_is_zero(self):
+        duration = datetime.timedelta(days=3)
+        self.mock_clock.now.return_value = datetime.datetime(
+            2000, 1, 1, 0, 0, 0, 0, tzinfo=pytz.utc)
+        timer = clock.Timer(self.mock_clock, duration)
+        timer.set_remaining(datetime.timedelta(seconds=0))
+        self.assertTrue(timer.expired())
+
+    def test_set_remaining_succeeds_when_remaining_equals_duration(self):
```
JeetShetty: This test will still pass if ```set_remaining()``` does nothing, not sure if that matters, though.

greenpithumb/greenpithumb.py
mtlynch: Ah! Good catch

tests/test_clock.py
mtlynch: True, fixed.

greenpithumb/clock.py
mtlynch: I want to be more defensive here because I feel like it's more likely that there's logic going on on the other side. For something like "make sure the parameter is not `None`" it's kind of a loud coding mistake on the caller's part. For something like this, the caller is probably isn't calling with something static like `timer.set_remaining(datetime.timedelta(minutes=5))`. They're probably doing some kind of arithmetic to calculate the time remaining.

greenpithumb/greenpithumb.py
```diff
@@ -139,6 +141,18 @@ def make_pump_manager(moisture_threshold, sleep_windows, raspberry_pi_io,
...
+    if last_pump_time:
+        logger.info('last watering was at %s', last_pump_time)
+        time_remaining = max(
+            datetime.timedelta(seconds=0),
+            (last_pump_time + pump_interval) - clock.Clock().now())
+    else:
+        logger.info('no previous watering found')
+        time_remaining = pump_interval
+    logger.info('max time until until next watering: %s', time_remaining)
+    pump_timer.set_remaining(time_remaining)
```
JeetShetty: (optional) I would move this to the ```if last_pump_time:``` conditional block and delete line 153

tests/test_clock.py
JeetShetty: Oh, I actually just meant that ```timer._end_time``` doesn't change when the duration is equal to ```time_remaining```. But, now that I type that out I realize that's because we're actually using the initial ```clock.now()``` return value twice, once when ```reset()``` is called upon initialization, and once when ```set_remaining()``` is called.

mtlynch:
>I just realized that if pump_interval is set to 7 days, and the program is restarted every 6 days, the pump will never run (assuming the user isn't also using soil moisture level to trigger the pump). Not sure if we want to address that in this PR, or create an issue and address it later, or if we just aren't worried about this at all.

Oh! Good catch. I think the sensible thing to do is to just water on next poll if there is no watering history. It makes more sense that way because if the user just started GreenPiThumb, we have to assume some prior state and it makes more sense to assume the plant needs to be watered immediately as opposed to the plant was *just* watered.

tests/test_clock.py
mtlynch: >But, now that I type that out I realize that's because we're actually using the initial clock.now() return value twice, once when reset() is called upon initialization, and once when set_remaining() is called.

Ah, I think that's actually a good thing. The test shouldn't know when `timer` is checking the clock, so if we only adjust it when control is outside of `timer`, it's clear to the reader of the test what time the timer thinks it is in each function call.

